PR #4, opened 2017-11-10
Unpack Bazel in the repository rule

damienmg:
This way Bazel is unpacked only once per version for all the tests.

### DIFF
--- a/bazel_integration_test/BUILD
+++ b/bazel_integration_test/BUILD
@@ -20,7 +20,7 @@ py_library(
     py_library(
         name = "python_version_" + v,
         srcs = [":python_generated_version_src_" + v],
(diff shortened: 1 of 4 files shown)


damienmg:
Seems like there is some issue on Linux... Seems like all timestamp are not correctly set, let me try something else.

damienmg:
This is due to this line: https://github.com/bazelbuild/bazel/blob/9135b7bfc4245e16bd6599e22dce0ba077e0d10a/src/main/cpp/blaze.cc#L1109, let me see if I can find a solution

damienmg:
Ok so Bazel set the install base mtime so cleanup script know which install base to delete. A solution  might be to symlink the content of the install base and set the install base to another dir.

damienmg:
Ok this is hacky but it worked :)

PR #5, opened 2017-11-10
Create a test suite per test macro

damienmg:


damienmg:
We could change the ci script to invoke those target specifically but that would be less adaptable to change (i.e. we would need to maintain that list).

Documentation of this repo is still TBD.

ittaiz:
I'm fiddling with a `bazel_java_integration_test` that runs `bazel_java_integration_test` and calls the suite :)
I'll probably stop soon and pick it up on Sunday.
You can merge if you want but I'd really like to get that test in.
as @johnynek said my testing bar is high. I'm probably uptight about testing as he is about isolation and reproducibility. We both care greatly about correctness in our own ways :)

ittaiz:
I'm close.
Can't push since I don't have write and would rather not fork at this time of night.
Pretty certain all I'm missing is to add the junit/hamcrest deps to the WORKSPACE I create in my test
[test_in_bazel_patch.txt](https://github.com/bazelbuild/bazel-integration-testing/files/1460725/test_in_bazel_patch.txt)

damienmg:
I just gave you write access, feel free to add the test. This PR is not urgent to merge.

ittaiz:
Thanks!
Now that I think of it adding the junit/hamcrest deps is not that easy since you depend on `@org_hamcrest_core//jar` which will be a bit cumbersome for me to simulate.
To use a maven_jar dependency in the test I'd need to start up a fake maven server which feels too much of an overkill (for this specific test).
Another alternative is to manually create a local repository called `org_hamcrest_core` and add a `jar` package there with a target called `jar`.
That also feels too much of work.
I think this correlates a lot to the discussion about canonical naming since I'd like to supply you with that dependency even though it's not from maven. My best alternative seems to use `bind` but I know it's on its way out.

WDYT?
I think this discussion is important to the usability of this library since it surfaces a common use-case of a rule testing itself.

ittaiz:
Ok,
We have a passing integration test which runs bazel with the integration testing rule itself :) Meta-test...

The amazing thing about tests is that they give so much feedback about the API and the needs.

Many things surfaced which I'll open separate issues about but just to clarify here as well:	
* Need to split about the logic from BazelBaseTestCase to WorkspaceDriver which will be agnostic from JUnit and BazelBaseTestCase will use it.

* Diagnostics
  * dump workspace layout on error (exists in contents method)
  * output logs of inner tests (exists in logsOfInternalTests method)

  * Exit code higher level matcher- maybe (exists in successfulExitCode, commandDescription)
       
  * Utility method for cat for file content to verify it contains what you expect- maybe (`Files.readAllLines(workspace.toPath().resolve("IntegrationTestSuiteTest.java")).forEach(System.out::println);`)

* Workspace setup
  * copyFromRunFiles entire directories
  * strip prefix from copyFromRunfiles
  * External repositories handling- maybe (exists in addExternalRepositoryFor, repositoryDeclarationFor, writeWorkspaceFileWithRepositories)

Essentially the "maybe"s are features I'm not sure that should be generalized and I'll wait for more feedback (from me or others) about the necessity.
The others are features I'll open issues for and start working on them (unless you'll have objections)

The only dillema I have about the current code is that tests take a lot of time to run and I think I'm not leveraging the unpacking. 
Can you help me with it?

ittaiz:
btw, I rebased on top of master to see if the unpacking solves the problem but it doesn't (or maybe I'm not utilizing it well)

ittaiz:
I think the cla problem might be from the rebase:
```
commit 42d2bd596cb8b015ed687fbeef3587779b3003f2
Author: Damien Martin-Guillerez <damienmg@users.noreply.github.com>
Date:   Thu Nov 9 23:37:43 2017 -0800

    Unpack Bazel in the repository rule (#4)
    
    This way Bazel is unpacked only once per version for all the tests.
```
Note the `damienmg@users.noreply.github.com`

damienmg:
humm ok let me fix that

javatests/build/bazel/tests/integration/BazelBaseTestCaseTest.java
```diff
@@ -14,19 +14,201 @@
...
+                "    name = \"" + testName + "\",\n" +
+                "    test_class = \"" + testName + "\",\n" +
+                "    srcs = [\"" + testName + ".java\"],\n" +
+                ")\n");
+    }
+
+    private void loadIntegrationTestRuleIntoWorkspace() throws IOException {
+        setupRuleSkylarkFiles();
+        setupRuleCode();
+        addExternalRepositoryFor("org_junit", "junit-4.11.jar");
```
damienmg: I would use wildcard and globs to avoid encoding the version name here.

javatests/build/bazel/tests/integration/BazelBaseTestCaseTest.java
```diff
@@ -14,19 +14,201 @@
...
+                .flatMap(Collection::stream);
+    }
+
+    private void setupPassingTest(final String testName) throws IOException {
+        writePassingTestJavaSource(testName);
+        writeTestBuildFile(testName);
+    }
+
+    private void writeTestBuildFile(final String testName) throws IOException {
+        scratchFile("BUILD", "load(\"//:bazel_integration_test.bzl\", \"bazel_java_integration_test\")\n" +
```
damienmg: You can use single quote to avoid backslash and `", "` instead of `\n" + "`, the scratchFile joins the lines

javatests/build/bazel/tests/integration/BazelBaseTestCaseTest.java
```diff
@@ -14,19 +14,201 @@
...
-/** {@link BazelBaseTestCase}Test */
+/**
+ * {@link BazelBaseTestCase}Test
+ */
+//suppress since same parameter value is ok for tests readability, tests should encapsulate and not hide
+@SuppressWarnings("SameParameterValue")
 public final class BazelBaseTestCaseTest extends BazelBaseTestCase {
 
-  @Test
-  public void testVersion() throws Exception {
```
damienmg: The style is not google java style :) I can do a pass with the formatter at the end of the review if you prefer.

javatests/build/bazel/tests/integration/BazelBaseTestCaseTest.java
```diff
@@ -14,19 +14,201 @@
...
 import static org.junit.Assert.assertEquals;
 
-import org.junit.Test;
-
-/** {@link BazelBaseTestCase}Test */
+/**
+ * {@link BazelBaseTestCase}Test
+ */
+//suppress since same parameter value is ok for tests readability, tests should encapsulate and not hide
+@SuppressWarnings("SameParameterValue")
```
damienmg: What is that Warning? Eclipse says it doesn't know it.

javatests/build/bazel/tests/integration/BazelBaseTestCaseTest.java
```diff
@@ -14,19 +14,201 @@
...
+        loadIntegrationTestRuleIntoWorkspace();
+        setupPassingTest("IntegrationTestSuiteTest");
+
+        Command cmd = bazel("test", "//:IntegrationTestSuiteTest");
+        final int exitCode = cmd.run();
+
+        org.hamcrest.MatcherAssert.assertThat(exitCode, is(successfulExitCode(cmd, workspace)));
+    }
+
+    private TypeSafeDiagnosingMatcher<Integer> successfulExitCode(final Command cmd, final File workspace) {
```
damienmg: Should that be a utility moved in the base test class?

javatests/build/bazel/tests/integration/BazelBaseTestCaseTest.java
ittaiz: It's an intellij one.
It's about having a parameter with only one usage so it suggests to inline the parameter.

javatests/build/bazel/tests/integration/BazelBaseTestCaseTest.java
ittaiz: no, I'll do it. 
If I'm going to contribute code here I should setup my system :)

javatests/build/bazel/tests/integration/BazelBaseTestCaseTest.java
ittaiz: So this is what I wrote above. Sorry it's a wall of text but it's important.
This one specifically is in the "maybe" section, I can move it up now but I'm not yet certain it's generic enough. If you think it is then maybe that's the second vote :)

damienmg: I am fine with not moving it now if you prefer (and do it in a follow-up change) but it is definitely a nice utilty you want to use for other stuff :)

ittaiz: I'll add it to the following PRs. Basically I have around 6-7 small follow up PRs I want to submit

javatests/build/bazel/tests/integration/BazelBaseTestCaseTest.java
ittaiz: done. Needed some tricks to concat lists and individual strings so take a peak to see if you like

javatests/build/bazel/tests/integration/BazelBaseTestCaseTest.java
ittaiz: I agree. Given copyFromRunfiles only copies specific files and doesn't currently supports directories or regular expressions I'd like to wait until I add directory copying which will make this change much easier.
WDYT?

javatests/build/bazel/tests/integration/BazelBaseTestCaseTest.java
ittaiz: done, I ran it on both java files.
Funny enough it changed some of your own code :)

ittaiz:
Ok, I've done all the changes. Waiting for your thoughts about the comments I left.
Also if there is anything we can do about the performance that would be great

ittaiz:
@damienmg friendly ping on this PR and also a small request:
Can you enable github issues? I want to open issues for the things we talked about here

tools/BUILD
```diff
@@ -5,3 +5,7 @@ py_binary(
     srcs = ["create_bazel_hashdict.py"],
     deps = ["@com_google_python_gflags//:gflags"],
 )
+filegroup(
+    name = "tools",
+    srcs = glob(["*"])
+)
```
damienmg: Add ending new line

ittaiz: Done

damienmg:
@googlebot: I signed it *trying to retrigger the test*

PR #6, opened 2017-11-14
Supporting no srcs use-case

ittaiz:
In rules_scala we want to use the java rule without srcs so that we can write the tests in scala.
There is currently a problem with `deps` and no srcs which I fixed.
This is based on top of #5 but I'll rebase if need be.
Only last commit is relevant


damienmg:
Can you rebase it?

tools/BUILD
```diff
@@ -5,3 +5,7 @@ py_binary(
     srcs = ["create_bazel_hashdict.py"],
     deps = ["@com_google_python_gflags//:gflags"],
 )
+filegroup(
+    name = "tools",
+    srcs = glob(["*"])
+)
```
damienmg: Missing new line

javatests/build/bazel/tests/integration/BUILD
```diff
@@ -6,3 +6,15 @@ bazel_java_integration_test(
...
+java_library(
+    name = "BazelIntegrationTestNoSourcesSrc",
+    srcs = ["BazelIntegrationTestNoSourcesTest.java"],
+    deps = ["@org_junit//jar"] 
+)
+
+bazel_java_integration_test(
+    name = "BazelIntegrationTestNoSources",
+    runtime_deps = [":BazelIntegrationTestNoSourcesSrc"],
+    test_class = "build.bazel.tests.integration.BazelIntegrationTestNoSourcesTest",
```
damienmg: Name the test target BazelIntegationTestNoSourcesTest so we can remove the "test_class"

ittaiz: This was actually on purpose since that's how we're going to use it on rules_scala.
Do you still want me to change it?

ittaiz: wdyt?

ittaiz:
Ping

ittaiz:
@damienmg any chance you can give a quick look and sign off? Since you started I'd rather not add @or-shachar to this PR

javatests/build/bazel/tests/integration/BUILD
damienmg: I don't think I understand the purpose. I am fine keeping it that way, but I'd like to have a comment with the reason in that case so we don't try to rename it.

damienmg:
I just have a detail left so LGTM :)

ittaiz:
added the comment.
Do you want to review it or should I merge?

javatests/build/bazel/tests/integration/BUILD
```diff
@@ -6,3 +6,17 @@ bazel_java_integration_test(
...
     data = ["//:root_bzl", "//tools:tools"]
 )
+
+java_library(
+    name = "BazelIntegrationTestNoSourcesSrc",
+    srcs = ["BazelIntegrationTestNoSourcesTest.java"],
+    deps = ["@org_junit//jar"] 
+)
+
+#the below uses test_class instead of renaming the BazelIntegrationTestNoSourcesSrc target
```
damienmg: nit: white space after # (line below too) + add caps.

Also ending point.

ittaiz: done

PR #7, opened 2017-11-15
Add golang support

achew22:
Add support for testing in a new language, Go!


go/bazel_integration_test.bzl
```diff
@@ -0,0 +1,50 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# Python integration test framework for using Bazel
+load("//tools:common.bzl", "BAZEL_VERSIONS")
+load("//tools:repositories.bzl", "bazel_binaries")
+load("@io_bazel_rules_go//go:def.bzl", "go_test")
+
+def _make_compatable_version_list():
```
damienmg: compat*i*ble here and below

achew22:
PTAL

WORKSPACE
```diff
@@ -22,9 +23,39 @@ py_library(
...
+## Java
+
 maven_jar(
     name = "com_google_truth",
     artifact = "com.google.truth:truth:jar:0.31",
 )
+
+## golang
+
+git_repository(
```
jayconrod: Consider using a newer version with `http_archive` instead:

```
http_archive(
    name = "io_bazel_rules_go",
    url = "https://github.com/bazelbuild/rules_go/releases/download/0.7.0/rules_go-0.7.0.tar.gz",
    sha256 = "91fca9cf860a1476abdc185a5f675b641b60d3acf0596679a27b580af60bf19c",
)
```

BUILD
```diff
@@ -0,0 +1,8 @@
+load("@io_bazel_rules_go//go:def.bzl", "go_prefix", "gazelle")
+
+go_prefix("github.com/bazelbuild/bazel-integration-testing")
```
jayconrod: `go_prefix` is deprecated. You shouldn't need it if you have `importpath` for all Go rules in the subtree below.

go/bazel.go
```diff
@@ -0,0 +1,158 @@
...
+	"strings"
+	"syscall"
+)
+
+var _ = fmt.Printf
+
+var bazelVersion = "bazel_version was not set. This is an error."
+
+type TestingBazel interface {
+	ScratchDir(name string) error
```
jayconrod: Why provide these methods in a public interface? Seems like they would always have the same implementation, and they don't depend on any state. Can they be functions instead? Possibly private, if they're only used in bazel_test.go?

go/bazel.go
```diff
@@ -0,0 +1,158 @@
...
+	"path/filepath"
+	"strings"
+	"syscall"
+)
+
+var _ = fmt.Printf
+
+var bazelVersion = "bazel_version was not set. This is an error."
+
+type TestingBazel interface {
```
jayconrod: Do you expect other implementations of this interface? Might be simpler to just have the testingBazel struct be public if not.

go/bazel.go
```diff
@@ -0,0 +1,158 @@
...
+	return nil
+}
+
+// GetBazel returns the path to the Bazel binary that is available.
+func (b *testingBazel) GetBazel() string {
+	return b.bazel
+}
+
+func (b *testingBazel) RunBazel(args []string) (int, string, string) {
+	find := exec.Command("ls", "-laR")
```
jayconrod: This looks like debug code. Can it be removed?

go/bazel.go
```diff
@@ -0,0 +1,158 @@
...
+	}
+
+	// When there isn't an error we know it exited successfully
+	return 0, stdout.String(), stderr.String()
+}
+
+// IsWindows returns true/false depending on if you're on Windows.
+func (b *testingBazel) IsWindows() bool {
+	// TODO: Move this into a linux/darwin specific file.
+	return false
```
jayconrod: `runtime.GOOS` will tell you this.

go/bazel.go
```diff
@@ -0,0 +1,158 @@
...
+}
+
+// Rlocation Returns the absolute path to a runfile.
+func (b *testingBazel) Rlocation(runfile string) string {
```
jayconrod: `github.com/bazelbuild/rules_go/go/tools/bazel.Runfile` should do the same thing.

go/BUILD
```diff
@@ -0,0 +1,19 @@
...
+load("//:bazel_integration_test.bzl", "bazel_go_integration_test")
+load("@io_bazel_rules_go//go:def.bzl", "go_library")
+
+bazel_go_integration_test(
+    name = "bazel_test",
+    srcs = [
+        "bazel_test.go",  # keep
+    ],
+    library = ":go_default_library",
+)
```
jayconrod: Need an `importpath` attribute here so that `go_prefix` can be deleted.

BUILD
achew22: Done

WORKSPACE
achew22: Done

go/BUILD
achew22: Done

go/bazel.go
achew22: I have removed the interface

go/bazel.go
achew22: I just removed IsWindows since I'm going to defer runfiles to `github.com/bazelbuild/rules_go/go/tools/bazel.Runfile`

go/bazel.go
achew22: Done

go/bazel.go
achew22: Done

go/bazel.go
achew22: These have been moved to be static methods. I attempted to make bazel_test.go effectively an external test. I have made that more formal in this commit.

achew22:
@jayconrod, I just uploaded a commit responding to your comments. Please take a look

BUILD
damienmg: Still here?

jayconrod:
@damienmg Sure I'm happy to co-own this. Feel free to send any more Go code reviews my way.

achew22:
Rebased/squashed commit and removed go_prefix. Thanks for the review!

BUILD
```diff
@@ -1,5 +1,11 @@
...
     srcs = glob(["*.bzl"]),
-    visibility = ["//visibility:public"]
-)
\ No newline at end of file
+    visibility = ["//visibility:public"],
+)
+
+gazelle(
+    name = "gazelle",
+)
```
jayconrod: You'll probably still want the `prefix` attribute in the `gazelle` rule. It's just the standalone `go_prefix` rule that's deprecated.

damienmg:
Gave you both write access for the go part. I let @jayconrod finish the review.

achew22:
@damienmg I had to make a few additional changes to the java source to stub out the golang rules. Could you take a look and rebase/merge if that is okay. Jay has already approved the other changes. This should leave the gocode isolated enough that I don't need to touch anything else for a while.

damienmg:
I see a lot of change from the baseline. (That's why I prefer a rebase workflow but GitHub is lacking to support it).

Is it possible to get a clean rebase of the change?

PR #8, opened 2017-11-20
Make workspace name match repo name

achew22:


achew22:
Hrm... that's a good question... My gut instinct is that it somewhere in the python code that is hardcoding the runfiles lookup to use the old workspace name, but the token doesn't appear in that file at all... very strange.

https://github.com/bazelbuild/bazel-integration-testing/blob/master/bazel_integration_test/test_base.py

ittaiz:
Silly me, I added the usage in my test, you might not have noticed it.
Change also
in javatests/build/bazel/tests/integration/BazelBaseTestCaseTest.java

achew22:
Ha! I read the errors report totally backward. I saw all the python entries at the bottom and assumed they were the ones that failed. Obviously that's not the case. Updating the java test is clearly the way to go.

achew22:
@damienmg I have rebased this so that I'm the only committer now. Should be good to go. In fact I was expecting the CLAbot to even retrigger

damienmg:
This is weird that @googlebot is not reporting... Let's try to close and reopen

PR #11, opened 2018-01-10
Parse the Bazel version as a tuple of ints instead of strings

jin:
This fixes the tuple comparison in `check_bazel_version`, as seen in the incorrect version comparisons in https://github.com/bazelbuild/bazel/issues/4425

### DIFF
--- a/bazel_version.bzl
+++ b/bazel_version.bzl
@@ -1,19 +1,16 @@
 # Copied from https://github.com/tensorflow/tensorflow/blob/dfa7d2d2953c3d33d2e8deb898240be008de5be6/tensorflow/workspace.bzl
 
 # Parse the bazel version string from `native.bazel_version`.
+# e.g.
+# "0.10.0rc1 abc123d" => (0, 10, 0)
+# "0.3.0" => (0, 3, 0)
 def _parse_bazel_version(bazel_version):
-  # Remove commit from version.
-  version = bazel_version.split(" ", 1)[0]
-
-  # Split into (release, date) parts and only return the release
-  # as a tuple of integers.
-  parts = version.split('-', 1)
-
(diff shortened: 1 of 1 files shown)


jin:
cc @ittaiz @achew22 @cgrushko FYI I'm also adding this file to @bazel_tools, so it'll be bundled with Bazel releases.

achew22:
@jin, will you be moving `_parse_bazel_version` or `check_bazel_version`? Also, would that be included in 0.10?

jin:
The entire file - there's a internal changelist pending to add the file to the Bazel source. I hope to get this in for 0.10.0, but it's going to be pointless because the file doesn't exist in prior versions' `@bazel_tools`. For now, existing users can depend on this repository for this file.

achew22:
On the plus side, if you try to load it in 0.9.0, it will fail and blow up which is the same effect but with a junkier error message :wink:

bazel_version.bzl
```diff
@@ -1,19 +1,16 @@
...
-  # Split into (release, date) parts and only return the release
-  # as a tuple of integers.
-  parts = version.split('-', 1)
-
-  # Turn "release" into a tuple of strings
-  version_tuple = ()
-  for number in parts[0].split('.'):
-    version_tuple += (str(number),)
-  return version_tuple
+  for i in range(len(bazel_version)):
```
katre: I feel like this should be broken out into a separate function.

jin:
As mentioned in https://github.com/bazelbuild/bazel/issues/4433#issuecomment-356741513, I think a better home for this function is bazelbuild/bazel-skylib. This allows older Bazel versions access to this function, and bazel-skylib contains functions used for writing build rules anyway. Will create another PR to get this file from Skylib once it's in.

PR #14, opened 2018-01-16
using try with resources to fix error prone failure

ittaiz:
fixes https://github.com/bazelbuild/bazel/issues/4453
@iirina I stumbled on to the above issue by chance. I'd really appreciate it If you could open an issue next time so I can take care of it :)


ittaiz:
@iirina @dslomov did the same change and asked me to review :) see #15

PR #22, opened 2018-02-26
support running tools/bazel script in java integration tests

natansil:
@damienmg @ittaiz please review.
The local tests pass and also the tests that rely on this change also pass

### DIFF
--- a/tools/bazel.sh
+++ b/tools/bazel.sh
@@ -18,5 +18,28 @@ if [ ! -d "${INSTALL_BASE}" ]; then
   done
 fi
 
-# Finally run Bazel with the new intall base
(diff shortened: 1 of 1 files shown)


ittaiz:
@philwo any idea why the build is stuck for a few good hours?

philwo:
Jenkins is probably already half dead by now - all we could do is reboot some VMs and hope for the best, but...

@buchgr Didn’t we enable presubmit checks for this repo? At least for Linux it should be fine, right, even if it takes a long time to test?

ittaiz:
Why should it take a long time? :(

ittaiz:
I'm merging after I saw it pass locally.
@philwo this is the second time I had to do this... We really need to solve this

PR #31, opened 2018-03-01
support running tools/bazel script in java integration tests

natansil:


natansil:
I suppose that the test failure relate to RBE.
for instance:
`WORKSPACE_DIR="${PWD}"` maybe problematic in RBE

the logs don't state anything about what causes the failure to run bazel...

ittaiz:
@philwo @dslomov we would really appreciate your help on this, we've been trying for a few days and are pretty much stuck.

What we're trying to do is to make the script running bazel in `bazel-integration-testing` more like the actual script (in that it delegates to tools/bazel if one exists).

We do it because our current implementation of WORKSPACE.resolved adds our own bazel script under tools/bazel for the workspace resolution.

There is an additional challenge in the current setup (pre our PR) in which there is some `INSTALL_BASE` fiddling, done so that two different test cases using the same bazel version won't need to pay the cost of extracting bazel twice (it's 10-15 seconds...).

One last note is that it's passing locally and failing on buildkite (@philwo does buildkite use remote execution or only remote cache? if only remote cache is there a way to run the same container locally?)

philwo:
Hi @ittaiz,

happy to help here, but I have no clue what "bazel-integration-testing" is or what it does. It's not something that Bazel itself uses or that I personally ever used. 😓

Will gladly look into any Bazel and CI infrastructure related areas though!

> One last note is that it's passing locally and failing on buildkite (@philwo does buildkite use remote execution or only remote cache? if only remote cache is there a way to run the same container locally?)

Buildkite is using remote caching only. What do you mean with "run the same container locally"? We're not using (Docker) containers on Buildkite.

Also, I don't know if that matters, but the Bazel that we install on the Buildkite VMs is currently what the normal installer would install (e.g. /usr/local/bin/bazel is actually that shell script that looks into tools/bazel first and stuff). I'm not yet sure if that's a good idea, or whether we should install just the plain binary into /usr/local/bin and not the wrapper script.

philwo:
Regarding the Buildkite VMs itself, you can find their setup scripts here: https://github.com/bazelbuild/continuous-integration/tree/master/buildkite

Check for example the [Buildkite Ubuntu 14.04 image](https://github.com/bazelbuild/continuous-integration/blob/master/buildkite/create_images.py#L46): In order to create the Buildkite VM image, we take the latest official Google Cloud Ubuntu 14.04 image and then run a number of shell scripts on the VM (e.g. to install Azul Zulu, Bazel, Android SDK, ...). When it all passed without errors, we shutdown the VM, snapshot the disk and use it as the image for all Buildkite Ubuntu 14.04 workers.

The scripts don't do anything magic and I think you should be easily able to just copy & paste the relevant ones. :)

natansil:
@philwo I eventually traced the problem to shell test statement syntax. In order for the tests not to fail I've changed the statement from 
`if [[ -e "${WORKSPACE_DIR}/WORKSPACE" ]]; then`
to
`if [ -e "${WORKSPACE_DIR}/WORKSPACE" ]; then
`
Does the image only support POSIX shell, not bash?

natansil:
@ittaiz please review

philwo:
It’s a standard Ubuntu - I wouldn’t be surprised if /bin/sh is not bash but dash:

https://wiki.ubuntu.com/DashAsBinSh

I would recommend to use #!/bin/bash and not #!/bin/sh in scripts, or alternatively consciously write POSIX-only shell scripts withou bashisms, which is good for portability (but Bazel itself requires bash, so...).

tools/bazel.sh
```diff
@@ -18,5 +18,28 @@ if [ ! -d "${INSTALL_BASE}" ]; then
   done
 fi
 
-# Finally run Bazel with the new intall base
-"${BASEDIR}/bin/bazel-real" --install_base="${INSTALL_BASE}" "$@"
+export BAZEL_REAL="${BASEDIR}/bin/bazel-real"
+
```
philwo: Collapse vertical whitespace to just one free line between blocks?

tools/bazel.sh
```diff
@@ -18,5 +18,28 @@ if [ ! -d "${INSTALL_BASE}" ]; then
...
+WORKSPACE_DIR="${PWD}"
+while [ "${WORKSPACE_DIR}" != / ]; do
+    if [ -e "${WORKSPACE_DIR}/WORKSPACE" ]; then
+      break;
+    fi
+    WORKSPACE_DIR="$(dirname "${WORKSPACE_DIR}")"
+done
+readonly WORKSPACE_DIR
+
```
philwo: Collapse vertical whitespace.

tools/bazel.sh
```diff
@@ -18,5 +18,28 @@ if [ ! -d "${INSTALL_BASE}" ]; then
...
+  readonly WRAPPER="${WORKSPACE_DIR}/tools/bazel"
+
+  if [ -x "${WRAPPER}" ]; then
+       export INSTALL_BASE="${INSTALL_BASE}"
+    "${WRAPPER}" "$@"
+  else
+       "${BASEDIR}/bin/bazel-real" --install_base="${INSTALL_BASE}" "$@"
+  fi
+
+fi
```
philwo: Add newline at end of file.

tools/bazel.sh
```diff
@@ -18,5 +18,28 @@ if [ ! -d "${INSTALL_BASE}" ]; then
...
+done
+readonly WORKSPACE_DIR
+
+
+
+if [ -e "${WORKSPACE_DIR}/WORKSPACE" ]; then
+  readonly WRAPPER="${WORKSPACE_DIR}/tools/bazel"
+
+  if [ -x "${WRAPPER}" ]; then
+       export INSTALL_BASE="${INSTALL_BASE}"
```
philwo: Indentation is not consistent here.

tools/bazel.sh
```diff
@@ -18,5 +18,28 @@ if [ ! -d "${INSTALL_BASE}" ]; then
...
+    "${WRAPPER}" "$@"
+  else
+       "${BASEDIR}/bin/bazel-real" --install_base="${INSTALL_BASE}" "$@"
```
philwo: Indentation.

tools/bazel.sh
```diff
@@ -18,5 +18,28 @@ if [ ! -d "${INSTALL_BASE}" ]; then
   done
 fi
 
-# Finally run Bazel with the new intall base
-"${BASEDIR}/bin/bazel-real" --install_base="${INSTALL_BASE}" "$@"
+export BAZEL_REAL="${BASEDIR}/bin/bazel-real"
```
philwo: This variable is unused?

tools/bazel.sh
```diff
@@ -18,5 +18,28 @@ if [ ! -d "${INSTALL_BASE}" ]; then
...
+
+
+WORKSPACE_DIR="${PWD}"
+while [ "${WORKSPACE_DIR}" != / ]; do
+    if [ -e "${WORKSPACE_DIR}/WORKSPACE" ]; then
+      break;
+    fi
+    WORKSPACE_DIR="$(dirname "${WORKSPACE_DIR}")"
+done
+readonly WORKSPACE_DIR
```
philwo: What does the "readonly" protect against in a ~40 line shell script that fits completely on one screen? :)

natansil: This is just to be consistent with [production bazel script](https://github.com/bazelbuild/bazel/blob/7be6ec66422e23705a8d8345675ee4c15bad3882/scripts/packages/bazel.sh#L73)

tools/bazel.sh
natansil: It's used externally, but I've also added an internal use

natansil:
@ittaiz pushed the changes according to @philwo's comments
Please merge

PR #36, opened 2018-03-26
scratchFile_Should_CreateFile_and_workspaceContents_ContainTheFile

maximn:
* Added com.spotify:hamcrest-optional

PR #37, opened 2018-03-26
Added .project(bazel) and IDEA to .gitignore

maximn:


damienmg:
Shouldn't that be in user specific .gitignore?

.gitignore
```diff
@@ -1 +1,4 @@
 /bazel-*
+/.project
+
```
ittaiz: can you please remove this empty line?

ittaiz:
good question. Bazel's own .gitignore file adds these and much more.
.project for example is something which is unrelated to @maximn as a user but is a by-product of some of the code of this lib

damienmg:
Ok make sense, maybe add comment to each section, e.g.:

```
# Bazel output directories
bazel-*
# IDE generated files
/idea
/.project
```
?

ittaiz:
update- actually .project seems like bazel-intellij related but still both are part of Bazel's own gitignore.
@damienmg wdyt?

damienmg:
As it make sense to have that as a default ignore to everybody that's fine by me.

ittaiz:
sounds good. only comment is that the bazel one needs to be `/bazel-*` in case a sub directory is called `bazel-foo`

damienmg:
Commit LGTM, there is just a missing newline at the end.

ittaiz:
@philwo I think we are at somewhat of an impasse with respect to PRs and maybe you have relevant inputs.
This PR started from master before #36 was merged and so I saw that I *have* to "update from master" as this is required.
On the other hand when I do so I get a failure from the cla bot :(
WDYT?

maximn:
added the new line.  @damienmg Why is it needed?

philwo:
@maximn Just for style guide reasons. We take care in Bazel's code to not have trailing whitespace in lines and that each file ends with a newline.

ittaiz:
@philwo interesting to see if asking pull request author to click the button can prevent this from happening.
In any case I'd appreciate you checking with the cla-bot maintainers about this since it surfaces in many of the projects

PR #33, opened 2018-03-21
 spin-off a WorkspaceDriver from BazelBaseTestCase

natansil:

PR #41, opened 2018-03-26
Minimize dependencies

ittaiz:
@damienmg I did two things here as the two separate commits suggest:
1. I minimized the dependencies of `bazel_java_integration_test` rule itself thanks to @natansil's work in #33 which separated between the `WorkspaceDriver` and the `BazelBaseTestCase`. This means users don't have to use junit/hamcrest.
2. I noticed that each test target has a data dependency on all of the bazel installers so //test/foo/bazel0.8.0 has a data dependency on all of the installers. AFAIU this can have an impact on remote execution since they will all need to be sent even though they're not needed, right? 
I moved the dependency into the rule itself instead of the library.
This meant opening a small backdoor for one of our own tests but I think the tradeoff is relevant.

WDYT?

### DIFF
--- a/tools/bazel_java_integration_test.bzl
+++ b/tools/bazel_java_integration_test.bzl
@@ -74,9 +74,7 @@ def bazel_java_integration_test(name, srcs=[], deps=None, runtime_deps=[],
   if not test_class:
     test_class = "%s.%s" % (_java_package(), name)
   add_deps = [
-    str(Label("//java/build/bazel/tests/integration")),
-    "@org_hamcrest_core//jar",
-    "@org_junit//jar",
+    str(Label("//java/build/bazel/tests/integration:workspace_driver")),
   ]
   if srcs:
     deps = (deps or []) + add_deps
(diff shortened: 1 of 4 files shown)


damienmg:
Btw, I don't see mac testing, we don't have macs with buildkite?

natansil:
why does our own test need to avoid getting the bazel-binaries?

ittaiz:
Indeed strange. @philwo can we have Mac builds?
@natansil
They need to avoid it since there I couldn’t find an easy way (after 2-3 hours of debugging) to add the entire bazel binary repository to the runfiles (which is needed if the bzl adds it as a data dependency)

ittaiz:
@damienmg done with the slash.
Is there an online google linter bot we can hook up here?

tools/bazel_java_integration_test.bzl
```diff
@@ -60,8 +60,10 @@ def _java_package():
   return ".".join(segments[idx+1:])
 
 
-def bazel_java_integration_test(name, srcs=[], deps=None, runtime_deps=[],
+def bazel_java_integration_test(name, srcs=[], deps=None, runtime_deps=[], data=[],
                                 jvm_flags=[], test_class=None,
+                                #flag to allow bazel_integration_testing own tests to work
```
damienmg: It is still missing a space here.

damienmg:
There is https://github.com/google/google-java-format but that's not really online and that's only for Java.

PR #43, opened 2018-03-27
implemented `copyDirectoryFromRunfiles `

maximn:


maximn:
@ittaiz

java/build/bazel/tests/integration/BazelBaseTestCase.java
```diff
@@ -95,6 +89,13 @@ protected void copyFromRunfiles(String path) throws IOException {
     driver.copyFromRunfiles(path);
   }
 
+  /**
+   * Copy the whole directory from the runfiles under {@code rootDirectoryPath} to the current workspace.
+   */
+  protected void copyFromRunfilesDirectory(String rootDirectoryPath) throws IOException {
```
ittaiz: wdyt about `copyDirectoryFromRunfiles`?

java/build/bazel/tests/integration/WorkspaceDriver.java
```diff
@@ -18,15 +18,11 @@
...
-import java.util.Arrays;
-import java.util.HashMap;
-import java.util.Map;
-import java.util.List;
-import java.util.ArrayList;
-import java.util.Collections;
-import java.util.stream.Stream;
 import java.nio.file.Path;
+import java.nio.file.Paths;
+import java.util.*;
```
ittaiz: https://google.github.io/styleguide/javaguide.html#s3.3.1-wildcard-imports forbids wildcard imports

java/build/bazel/tests/integration/WorkspaceDriver.java
```diff
@@ -155,6 +151,27 @@ protected void copyFromRunfiles(String path, String destpath) throws IOException
     Files.copy(origin.toPath(), dest.toPath());
   }
 
+  /**
+   * Copy the whole directory from the runfiles under {@code rootDirectoryPath} to the current workspace.
+   */
+  protected void copyFromRunfilesDirectory(final String rootDirectoryPath) throws IOException {
+    File root = getRunfile(rootDirectoryPath);
+    try (Stream<Path> paths = Files.walk(Paths.get(root.toURI()))) {
+      paths.filter(path -> new File(path.toUri()).isFile())
```
ittaiz: the java 8 way to do `new File(path.toUri()).isFile()` is with `java.nio.file.Files.isRegularFile`

java/build/bazel/tests/integration/WorkspaceDriver.java
```diff
@@ -155,6 +151,27 @@ protected void copyFromRunfiles(String path, String destpath) throws IOException
...
+              .forEach(file -> {
+                String relativeToRunfiles = file.toAbsolutePath().toString().substring(runfileDirectory.getPath().length());
+                Path relativeToRunfilesPath = Paths.get(relativeToRunfiles);
+                String destination = relativeToRunfilesPath.subpath(1, relativeToRunfilesPath.getNameCount()).toString();
```
ittaiz: wdyt about doing a private method here? for readability. This is probably something like "strip first directory". First is meh name but root is also problematic since it's root in relation to the runfiles

java/build/bazel/tests/integration/WorkspaceDriver.java
```diff
@@ -155,6 +151,27 @@ protected void copyFromRunfiles(String path, String destpath) throws IOException
...
 
+  /**
+   * Copy the whole directory from the runfiles under {@code rootDirectoryPath} to the current workspace.
+   */
+  protected void copyFromRunfilesDirectory(final String rootDirectoryPath) throws IOException {
+    File root = getRunfile(rootDirectoryPath);
+    try (Stream<Path> paths = Files.walk(Paths.get(root.toURI()))) {
+      paths.filter(path -> new File(path.toUri()).isFile())
+              .forEach(file -> {
+                String relativeToRunfiles = file.toAbsolutePath().toString().substring(runfileDirectory.getPath().length());
```
ittaiz: can you take a look at `Path#relativize`? I don't think it's what you need but it's definitely a neighbor :)
something with `file` and `runFilesDirectory`

java/build/bazel/tests/integration/WorkspaceDriver.java
ittaiz: why do you need to filter out directories? what happens if you don't? this probably makes sense if you're going to forbit it in the "regular" copyFromRunfiles method

java/build/bazel/tests/integration/WorkspaceDriver.java
maximn: Actually, it's exactly `relativize`

ittaiz:
The test passes with it? Since I seem to remember it’s different but if it
works then I guess I’m wrong

java/build/bazel/tests/integration/WorkspaceDriver.java
maximn: Because `copyFromRunfiles` shouldn't work for directories, we'll decide how to handle it later.

Also, `walk` do the traversal depth first so I'll get to the directories after the leafs meaning that the directory already exists.

maximn:
Yes, of course passed :)

java/build/bazel/tests/integration/WorkspaceDriver.java
ittaiz: Still wildcard here

java/build/bazel/tests/integration/WorkspaceDriver.java
```diff
@@ -155,6 +151,29 @@ protected void copyFromRunfiles(String path, String destpath) throws IOException
     Files.copy(origin.toPath(), dest.toPath());
   }
 
+  /**
+   * Copy the whole directory from the runfiles under {@code directoryToCopy} to the current workspace.
+   */
+  protected void copyDirectoryFromRunfiles(final String directoryToCopy) throws IOException {
+    File startingDirectory = getRunfile(directoryToCopy);
```
ittaiz: Should we handle the case where we get a file and and not a directory?

java/build/bazel/tests/integration/WorkspaceDriver.java
```diff
@@ -152,26 +152,28 @@ protected void copyFromRunfiles(String path, String destpath) throws IOException
...
-                  copyFromRunfiles(relativeToRunfiles, destination);
-                } catch (IOException e) {
-                  throw new RuntimeException(e);
-                }
-              });
+  protected void copyDirectoryFromRunfiles(final String directoryToCopy) throws IOException {
+    File startingDirectory = getRunfile(directoryToCopy);
+    try (Stream<Path> paths = Files.walk(startingDirectory.toPath())) {
+      paths.filter(path -> Files.isRegularFile(path))
+            .forEach(file -> {
```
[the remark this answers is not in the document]
maximn: Why do you think that it's expensive? The alternative is `Paths.get(runfileDirectory.toUri())` don't think that it matter a lot. Why are you concerned about performance here?

java/build/bazel/tests/integration/WorkspaceDriver.java
```diff
@@ -152,26 +152,28 @@ protected void copyFromRunfiles(String path, String destpath) throws IOException
...
-                String destination = relativeToRunfilesPath.subpath(1, relativeToRunfilesPath.getNameCount()).toString();
-
-                try {
-                  copyFromRunfiles(relativeToRunfiles, destination);
-                } catch (IOException e) {
-                  throw new RuntimeException(e);
-                }
-              });
+  protected void copyDirectoryFromRunfiles(final String directoryToCopy) throws IOException {
+    File startingDirectory = getRunfile(directoryToCopy);
```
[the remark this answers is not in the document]
maximn: Yes, I was trying to add such test but had some difficulties, let's talk about it f2f tomorrow.

java/build/bazel/tests/integration/WorkspaceDriver.java
maximn: Fixed the wildcard, IDEA is stubborn ;-)

maximn:
@ittaiz look at commit b36d830 , many method signatures now changed to declare that they thorws `BazelWorkspaceDriverException`. WDYT about changing it to a **non-checked exception**?

PR #49, opened 2018-03-28
Clean up bazel base test case proxy methods

maximn:


maximn:
@ittaiz

java/build/bazel/tests/integration/BazelBaseTestCase.java
```diff
@@ -17,13 +17,8 @@
 import org.junit.Before;
 import org.junit.BeforeClass;
 
-import java.io.File;
 import java.io.IOException;
-import java.util.List;
 
-/**
```
ittaiz: Why did you remove this? I think the comment still holds

maximn: `A base class to do integration test that call Bazel` - I don't feel like it adds useful information and it doesn't really call Bazel anymore.

maximn:
regarding  “contents” -> “workspaceDirectoryContents”
Let's talk about the name f2f. Also if we change it let's consider changing the signature to return `List<Path>` instead of `List<String>`

java/build/bazel/tests/integration/BazelBaseTestCase.java
ittaiz: I think you’re misreading it. It’s a base class to do an integration test which calls bazel. The base doesn’t call it. The test calls it.

maximn: Do you see value in this comment? Is it just project's style?

ittaiz:
Just project’s style.

PR #50, opened 2018-03-28
Command improvements

maximn:


maximn:
@ittaiz

ittaiz:
@damienmg @maximn and I talked f2f and we're actually going to change the code to read the stdout/stderr after the command finishes and be done with it.
Pro: Much simpler code
Con: One cannot write a test which runs `bazel` on one thread and query the logs on a different thread.

I feel the con is weak enough to do this change and worst case support that feature further down the road if anyone really wants it.
WDYT?

damienmg:
You might run into buffering issue when a command generates a lot of output. I don't remember the whole details but that was the reason it had to be a separate thread.

maximn:
@damienmg , I've read about this a bit now and you're right.

> Because some native platforms only provide limited buffer size for standard input and output streams, failure to promptly write the input stream or read the output stream of the subprocess may cause the subprocess to block, or even deadlock. 

[[java.lang.Process](https://docs.oracle.com/javase/8/docs/api/java/lang/Process.html)]

I'll remove the last commit.

java/build/bazel/tests/integration/Command.java
```diff
@@ -14,33 +14,35 @@
...
 import java.util.Collections;
+import java.util.LinkedList;
+import java.util.List;
 import java.util.Objects;
 
 
 /**
  * A utility class to spawn a command and get its output.
- *
+ * <p>
```
damienmg: nit: remove extraneous <p>

Issue #52, opened 2018-03-29
Rename WorkspaceDriver#contents to something that resonates more with directory contents/layout


ittaiz:
cc @maximn

ittaiz:
fixed by #53

PR #53, opened 2018-03-29
changed contents->workspaceDirectoryContents and return type to Path

maximn:


maximn:
@ittaiz  https://github.com/bazelbuild/bazel-integration-testing/issues/52

Issue #44, opened 2018-03-28
Move APIs from `java.io.File` to `java.nio.file.Path`


ittaiz:
`Path` is a newer api which also allows plugging in in memory filesystem for testing and in general is more recommended

ittaiz:
cc @maximn

maximn:
This is a breaking change. Did you mean to do that or just change the internals?

ittaiz:
Yes I did.
Cc @damienmg any strong objections?

damienmg:
No objection Path is definitely better.

ittaiz:
closed by #56

PR #56, opened 2018-03-29
Workspace driver file path

maximn:


maximn:
@ittaiz

maximn:
https://github.com/bazelbuild/bazel-integration-testing/issues/44

java/build/bazel/tests/integration/WorkspaceDriver.java
```diff
@@ -80,20 +78,20 @@ private static void unpackBazel(String version)
...
 
   private static Command prepareUnpackBazelCommand(String version)
-    throws  IOException {
+          throws IOException {
     List<String> command = new ArrayList<String>(Arrays.asList(
-        bazelVersions.get(version).getCanonicalPath(),
-        "--output_user_root=" + tmp, "--nomaster_bazelrc",
-        "--max_idle_secs=30", "--bazelrc=/dev/null",
-        "help"));
+            bazelVersions.get(version).toString(),
```
ittaiz: Do we want to normalize the path? Like get canonical? https://stackoverflow.com/questions/21552901/getting-canonical-path-from-path-in-java

java/build/bazel/tests/integration/WorkspaceDriver.java
```diff
@@ -31,46 +31,44 @@
...
   public static void setUpClass() throws IOException {
-    // Get tempdir
-    String _tmp = System.getenv("TEST_TMPDIR");
-    if (_tmp == null) {
-      File p = Files.createTempDirectory("e4b-tests").toFile();
-      p.deleteOnExit();
-      tmp = p;
+    String environmentTempDirectory = System.getenv("TEST_TMPDIR");
+    if (environmentTempDirectory == null) {
+      File tempDirectory = Files.createTempDirectory("e4b-tests").toFile();
```
ittaiz: Really nit but you do toFile and then afterwards toPath. Maybe save the first and then you can skip the second transformation?

java/build/bazel/tests/integration/WorkspaceDriver.java
```diff
@@ -135,54 +133,58 @@ public Command runBazelInDirectory(Path relativeDir, String... args) throws IOEx
...
     }
 
     List<String> command = new ArrayList<String>(Arrays.asList(
-      currentBazel.getCanonicalPath(),
-      "--output_user_root=" + tmp,
-      "--nomaster_bazelrc",
-      "--max_idle_secs=10",
-      "--bazelrc=/dev/null"));
-    for (String arg: args) {
+            currentBazel.toString(),
```
ittaiz: Same here about normalize

java/build/bazel/tests/integration/WorkspaceDriver.java
```diff
@@ -135,54 +133,58 @@ public Command runBazelInDirectory(Path relativeDir, String... args) throws IOEx
...
+    if (!Files.isDirectory(startingDirectory))
       throw new BazelWorkspaceDriverException("directoryToCopy MUST be a directory");
 
     if (!directoryToCopy.startsWith(stripPrefix))
       throw new BazelWorkspaceDriverException("The `stripPrefix` MUST be a prefix of `directoryToCopy`");
 
     Path stripPrefixPath = Paths.get(stripPrefix);
-    Path runfileDirectoryPath = runfileDirectory.toPath();
-    try (Stream<Path> paths = Files.walk(startingDirectory.toPath())) {
+    Path runfileDirectoryPath = runfileDirectory;
```
ittaiz: Can we drop this variable?

java/build/bazel/tests/integration/WorkspaceDriver.java
```diff
@@ -223,21 +225,18 @@ public void scratchExecutableFile(String path, Iterable<String> content) throws
...
   private File writeToFile(String path, Iterable<String> content) throws IOException {
-    File dest = new File(workspace, path);
-    if (!dest.getParentFile().exists()) {
-      dest.getParentFile().mkdirs();
-    }
-    Files.write(dest.toPath(), String.join("\n", content).getBytes(StandardCharsets.UTF_8));
-    return dest;
+    Path dest = createParentDirectoryIfNotExists(path);
+    Files.write(dest, String.join("\n", content).getBytes(StandardCharsets.UTF_8));
+    return dest.toFile();
```
ittaiz: Why do we need to do “toFile” here?

ittaiz: It’s because scratchExecutableFile needs to set it as executable.
I think you can use Files to do it on a path

java/build/bazel/tests/integration/WorkspaceDriver.java
maximn: I don't think so, from what I checked the way that we're generating the path provides the absolute path that we want here. No need to convert.

java/build/bazel/tests/integration/WorkspaceDriver.java
maximn: right

PR #57, opened 2018-03-29
Remove java.io.File

ittaiz:
Done from github UI, let’s see if build passes


ittaiz:
@damienmg can you take a look? It’s a really minor change

ittaiz:
Oh, then I’m sorry. I wasn’t aware.
Thanks!

PR #58, opened 2018-04-04
Add bazel RC support

anchlovi:
Current implementation uses `--bazelrc=/dev/null` which means that `bazel` will ignore `.bazelrc`. There are cases where this may block some use cases

CC: @natansil @ittaiz

### DIFF
--- a/java/build/bazel/tests/integration/WorkspaceDriver.java
+++ b/java/build/bazel/tests/integration/WorkspaceDriver.java
@@ -19,7 +19,13 @@ import java.nio.charset.StandardCharsets;
 import java.nio.file.Files;
 import java.nio.file.Path;
 import java.nio.file.Paths;
-import java.util.*;
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.Optional;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
 
--- a/java/build/bazel/tests/integration/WorkspaceDriver.java
+++ b/java/build/bazel/tests/integration/WorkspaceDriver.java
@@ -19,12 +19,7 @@ import java.nio.charset.StandardCharsets;
 import java.nio.file.Files;
 import java.nio.file.Path;
 import java.nio.file.Paths;
-import java.util.ArrayList;
-import java.util.Arrays;
-import java.util.Collections;
-import java.util.HashMap;
-import java.util.List;
-import java.util.Map;
+import java.util.*;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
 
@@ -114,13 +109,13 @@ public class WorkspaceDriver {
    * Prepare bazel for running, and return the {@link Command} object to run it.
    */
   public Command bazel(String... args) {
-    return bazel(null, args);
+    return bazel(Optional.empty(), args);
   }
 
   /**
    * Prepare bazel for running, and return the {@link Command} object to run it.
    */
-  public Command bazel(Path bazelrcFile, String... args) {
+  public Command bazel(Optional<Path> bazelrcFile, String... args) {
     return bazel(bazelrcFile, new ArrayList<>(Arrays.asList(args)));
   }
 
@@ -128,13 +123,13 @@ public class WorkspaceDriver {
    * Prepare bazel for running, and return the {@link Command} object to run it.
    */
   public Command bazel(Iterable<String> args) {
-    return bazel(null, args);
+    return bazel(Optional.empty(), args);
   }
 
   /**
    * Prepare bazel for running, and return the {@link Command} object to run it.
    */
-  public Command bazel(Path bazelrcFile, Iterable<String> args) {
+  public Command bazel(Optional<Path> bazelrcFile, Iterable<String> args) {
     return runBazelInDirectory(Paths.get(""), bazelrcFile, args);
   }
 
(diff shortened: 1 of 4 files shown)


java/build/bazel/tests/integration/WorkspaceDriver.java
```diff
@@ -113,33 +113,53 @@ public void setUp() throws IOException, InterruptedException {
...
+    return runBazelInDirectory(relativeToWorksapceDir, null, args);
+  }
+
+  public Command runBazelInDirectory(Path relativeToWorksapceDir, Path bazelrcFile, Iterable<String> args) {
     if (currentBazel == null) {
       throw new BazelWorkspaceDriverException("Cannot use bazel because no version was specified, "
               + "please call bazelVersion(version) before calling bazel(...).");
     }
 
+    String bazelRcPath = bazelrcFile == null ? "/dev/null" : workspace.resolve(bazelrcFile).toString();
```
ittaiz: Can you use java 8 Optional?

javatests/build/bazel/tests/integration/BazelBaseTestCaseTest.java
```diff
@@ -63,15 +63,30 @@ public void testRunningBazelInGivenRelativeDirectory() throws Exception {
...
 
     Command cmd = driver.bazel("test", "//:IntegrationTestSuiteTest");
     final int exitCode = cmd.run();
 
     org.hamcrest.MatcherAssert.assertThat(exitCode, is(successfulExitCode(cmd)));
   }
 
+  @Test
+  public void testUseBazelRcFile() throws Exception {
+    setUpTestSuit("IntegrationTestSuiteTest");
```
ittaiz: Why do we need here loadIntegrationTestRuleIntoWorkspace? I’d rather test the Driver functionality in the DriverTest

anchlovi: To test the driver I'll need to expose the `Command` internals (`args`)

ittaiz: Why do you think so? 
What I meant is exactly this test BUT instead of having the test you’re running be bazel_java_integration_testing you can have it be java_test.
You still run bazel and everything but on a much simpler case.
This test is complicated since it’s a meta test.
Your test AFAIU is a simpler one.
Does that make sense?

javatests/build/bazel/tests/integration/WorkspaceDriverIntegrationTest.java
```diff
@@ -0,0 +1,62 @@
...
+                "  @Test",
+                "  public void testSuccess() {",
+                "  }",
+                "}");
+    }
+
+    private List<String> workspaceWithJunit() {
+        return Arrays.asList(
+                "workspace(name = 'driver_integration_tests')",
+                "maven_jar(",
```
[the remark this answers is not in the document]
natansil: @ittaiz if what you suspect is the case, then is the alternative copying from runfiles?

java/build/bazel/tests/integration/WorkspaceDriver.java
```diff
@@ -113,33 +114,53 @@ public void setUp() throws IOException, InterruptedException {
   /**
    * Prepare bazel for running, and return the {@link Command} object to run it.
    */
-  public Command bazel(String... args) throws IOException {
-    return bazel(new ArrayList<>(Arrays.asList(args)));
+  public Command bazel(String... args) {
```
natansil: I think the area of setting up the bazel command is starting to become complex (7 methods already)
@anchlovi have you decided against the builder we talked about?

ittaiz: :) also thought to say builder but didn’t want to complicate his efforts even more.
OTOH now is the best time to break APIs as we learn them. If we decide on it then let’s loop in Damien as the other main user of this lib

javatests/build/bazel/tests/integration/WorkspaceDriverIntegrationTest.java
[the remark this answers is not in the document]
ittaiz: Yes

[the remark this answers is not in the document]
ittaiz: Shachar you can look at the history of BazelBaseTestCaseTest to see how we did it already. Natan can probably point towards the right place

java/build/bazel/tests/integration/WorkspaceDriver.java
ittaiz: I think that maybe it's better to defer the builder change to another PR since we should probably also discuss it with Damien. WDYT?

javatests/build/bazel/tests/integration/WorkspaceDriverIntegrationTest.java
```diff
@@ -0,0 +1,62 @@
...
+        driver.scratchFile("BUILD.bazel", testLabelNamed(testName));
+        driver.scratchFile("TestMe.java", passingTestNamed(testName));
+
+        driver.scratchFile(".bazelrc", "test --javacopt=\"-InvalidOpt\"");
+
+        Command cmd = driver.bazel(Optional.of(Paths.get(".bazelrc")), "test", "//:TestMe");
+        int returnCode = cmd.run();
+
+        assertNotEquals("bazel test return code", 0, returnCode);
+        assertTrue("stderr contains InvalidOpt failure", cmd.getErrorLines().stream().anyMatch(x -> x.contains("-InvalidOpt")));
```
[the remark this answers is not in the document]
anchlovi: Where?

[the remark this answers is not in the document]
ittaiz: x->x.contains...

ittaiz:
Btw,
This is indeed one of the improvements this library needs. Ease hermetic
dependency on external repositories.
I think there’s a bit of an easier way then what you did. I’ll try to play
around with it tomorrow.

PR #66, opened 2018-04-24
WorkspaceDriver: add a currentWorkspace()

damienmg:
The Eclipse Plugin test needs to inject the path of a workspace in the
command line flags to test its skylark aspect. Using
getCurrentWorkspace() it can knows the actual path of it.


ittaiz:
@damienmg wdyt about changing it to “currentWorkspace”? It’s a more fluent DSL than getters. I don’t feel strongly about it so you can merge as-is if you feel the getter is better

damienmg:
@ittaiz: I am fine with currentWorkspace. There is a mix of "getXxx" and "xxx" in that class would be good to be uniform at some point: #69.

PR #68, opened 2018-04-24
Simplify the BazelCommand API

hchauvin:


java/build/bazel/tests/integration/BuildEventCapturer.java
```diff
@@ -0,0 +1,72 @@
...
+  public BuildEventCapturer(Path file) {
+    this.file = file;
+  }
+
+  /**
+   * Gets the local file where the build events are stored.
+   */
+  Path file() throws IOException {
+    if (file == null) {
+      file = Files.createTempFile(
```
damienmg: Why not everything on one line?

java/build/bazel/tests/integration/BuildEventCapturer.java
```diff
@@ -0,0 +1,72 @@
+// Copyright 2017 The Bazel Authors. All rights reserved.
```
damienmg: 2017 -> 2018

java/build/bazel/tests/integration/WorkspaceDriver.java
```diff
@@ -228,6 +228,21 @@ public BazelCommand bazelCommand(List<String> args) {
...
+   */
+  public build.bazel.tests.integration.BazelCommand.Builder bazel(String arg, String... args) {
+    return bazel(
+        Stream.concat(Stream.of(arg), Stream.of(args)).collect(Collectors.toList()));
+  }
+
+  /**
+   * Returns a builder for invoking bazel.
+   */
+  public build.bazel.tests.integration.BazelCommand.Builder bazel(List<String> args) {
```
damienmg: here and above: there is no need for the full qualified name for BazelCommand, it is in the same package

tools/bazel_java_integration_test.bzl
```diff
@@ -13,6 +13,10 @@
 # limitations under the License.
 
 # Java integration test framework for using Bazel
+load(
+  "@bazel_tools//tools/build_defs/repo:http.bzl",
+  "http_archive",
+)
```
damienmg: This will break with older version of bazel.... maybe wrap it in a macro testing for native.http_archive?

java/build/bazel/tests/integration/WorkspaceDriver.java
hchauvin: There is a name clash with WorkspaceDriver.BazelCommand

tools/bazel_java_integration_test.bzl
hchauvin: If I'm not mistaken this was released in 0.4.2 (7 Dec 2016): https://github.com/bazelbuild/bazel/commit/451599a8d22d79a84e228c9c06207df490f02f9f

The oldest version of Bazel you test against here is 0.6.0.  What is the oldest version of Bazel you'd like to support?

java/build/bazel/tests/integration/BuildEventCapturer.java
hchauvin: see next commit

java/build/bazel/tests/integration/BuildEventCapturer.java
hchauvin: see next commit

java/build/bazel/tests/integration/WorkspaceDriver.java
damienmg: I prefer to not have 2 BazelCommand, I prefer to break the API there (which also relate to moving BazelCommand out of WorkspaceDriver)

tools/bazel_java_integration_test.bzl
damienmg: You are right, my bad. The bug in the early version won't be triggered so we are good to go.

hchauvin:
For the linting, are you talking only about the diff?  To be honest, I like a consistent style as well, but was not sure what it was by looking at the code.  I put a commit where I do a google-java-format everywhere, just in case, but I'm not sure this is the right way to proceed.

I think I should propose a lint.sh in another PR instead?  This would allow to have consistent formatting with the remaining bazel code base for BUILD and *.bzl files as well.

Concerning the extraction of the BazelCommand: I did not extract it, I added another incompatible one.  Should I remove the old one altogether and present a separate commit with only this BazelCommand?

java/build/bazel/tests/integration/WorkspaceDriver.java
hchauvin: Ok, I'll do that

damienmg:
It use to be google style but I did not followed the repository closely for some time and it seemed to have derived a little :)

For the linter that would be awesome, buildifier should be able to handle BUILD and bzl file. The best would be something that can be added a pre-commit hook :) (and that we can also run using travis CI).

My preference for the BazelCommand would be to keep only one of them and then we can have a clear diff of the API and discussed your proposed change.

java/build/bazel/tests/integration/WorkspaceDriver.java
```diff
@@ -31,16 +31,14 @@
...
 
-  private Path currentBazel = null;
+  Path currentBazel = null;
 
-  /**
-   * The current workspace.
-   */
-  private Path workspace = null;
+  /** The current workspace. */
+  Path workspace = null;
```
ittaiz: shared mutable state is a very bad idea...

hchauvin:
I agree, maybe it's better to focus on BazelCommand, I'll save the BEP for another time?

Concerning the formatting issue, I think it is actually better if #72 lands before, so that it becomes a non-issue?

damienmg:
sgtm

hchauvin:
Just rebased on top of the linter.

ittaiz:
@hchauvin this PR is now a suggestion to how you think the command object should be, right? We got linting out of the way and we're postponing the build event to after this one to ease review?

java/build/bazel/tests/integration/WorkspaceDriver.java
```diff
@@ -247,77 +252,17 @@ private BazelWorkspaceDriverException(String message) {
...
-
-  public BazelCommand bazelCommand(List<String> args) {
-    if (currentBazel == null) {
-      throw new BazelWorkspaceDriverException(
-          "Cannot use bazel because no version was specified, "
-              + "please call bazelVersion(version) before creating command(...).");
-    }
-    return new BazelCommand(args);
+  /** Returns a builder for invoking bazel. */
+  public BazelCommand.Builder bazel(String arg, String... args) {
```
damienmg: It seems like we are moving back and forth between "bazel" and "bazelCommand" for this function :) Anyway I prefer simply "bazel"

java/build/bazel/tests/integration/BUILD
```diff
@@ -4,11 +4,15 @@ java_library(
     name = "workspace_driver",
     testonly = True,
     srcs = [
+        "BazelCommand.java",
         "Command.java",
         "RepositoryCache.java",
         "WorkspaceDriver.java",
     ],
     visibility = ["//visibility:public"],
+    deps = [
```
damienmg: Is @Nullable worth adding a new dependency?

hchauvin: @Nullable is used only once here.  If I remember, it made sense when I added BEP support, but here it is used only once so I am remove it for now. See separate commit for better clarity.

java/build/bazel/tests/integration/BazelCommand.java
```diff
@@ -0,0 +1,148 @@
...
+
+    /** adds environment variable to execution runtime */
+    public Builder withEnvironmentVariable(String key, String value) {
+      this.environment.put(key, value);
+      return this;
+    }
+
+    private Command build() throws IOException {
+      String bazelRcPath =
+          bazelrcFile == null
```
ittaiz: why do you prefer null and trenary over optional?

java/build/bazel/tests/integration/BazelCommand.java
```diff
@@ -0,0 +1,148 @@
...
+          .withEnvironment(environment)
+          .build();
+    }
+
+    public BazelCommand run() throws IOException, InterruptedException {
+      Command cmd = build();
+      return new BazelCommand(cmd, args, cmd.run());
+    }
+
+    public BazelCommand mustRun() throws IOException, InterruptedException {
```
ittaiz: the phrasing `mustRun` confused me when I read the tests. WDYT about `mustRunSuccessfully`? longer but more revealing IMHO

javatests/build/bazel/tests/integration/BazelBaseTestCaseTest.java
```diff
@@ -54,21 +52,16 @@ public void testRunningBazelInGivenRelativeDirectory() throws Exception {
...
 
   @Test
   public void testTestSuiteExists() throws Exception {
     loadIntegrationTestRuleIntoWorkspace();
     setupPassingTest("IntegrationTestSuiteTest");
 
-    Command cmd = driver.bazelCommand("test", "//:IntegrationTestSuiteTest").build();
-    final int exitCode = cmd.run();
-
-    org.hamcrest.MatcherAssert.assertThat(exitCode, is(successfulExitCode(cmd)));
```
ittaiz: should we delete `successfulExitCode`?

ittaiz: it's important to note we're losing here:
workspaceDirectoryContents and logs of tests.
They were very valuable for me while developing but maybe we can live with them.
As a thought exercise (and maybe as input for us as maintainers of the lib) how should one write this test if they also want the above as diagnostics when it fails?

hchauvin: I delete successfulExitCode.

Logs of tests is something I wanted to be accessible through the Build Event Protocol interface.  In BazelCommand you have a toString() method which is used to describe the command in mustRun(), and it can be used in general for debugging.  Right now, the toString() method only gives the stderr and stdout, but I planned with the Build Event Protocol support to expand it to giving a list of produced artifacts and test logs.

I don't remember having in the workspaceDirectoryContents as well, so I just added it to toString() as well.

java/build/bazel/tests/integration/BazelCommand.java
hchauvin: The (official) position that I am following is that Optional in Java was not intended for class fields. Of course, this is a matter of opinion and coding style, and you are the final arbiter :)

Cf. answer by Brian Goetz: https://stackoverflow.com/questions/26327957/should-java-8-getters-return-optional-type/26328555#26328555 and also http://www.oracle.com/technetwork/articles/java/java8-optional-2175753.html: "The purpose of Optional is not to replace every single null reference in your codebase but rather to help design better APIs in which—just by reading the signature of a method—users can tell whether to expect an optional value."

Null is too much embedded in the Java language for using Optional in all cases.  Optional was primarily intended for method chaining and for public APIs, especially when they relate to streams.  IntelliJ agrees, and flags the use of Optional for fields.

It is also possible to use the @Nullable annotation to help findbugs finds spurious usage (findbugs is enabled by default in Bazel's java toolchain).

java/build/bazel/tests/integration/BazelCommand.java
hchauvin: That's derived from a convention in Golang.  That might be a bit foreign here, so I'll go with your suggestion.

hchauvin:
Concerning the feature envy: I see what you mean, but I do like the separation as well.

One way to solve that might be to get rid of Command.Builder and have a static method `Command Command.run(ProcessBuilder builder)` instead.  It would be responsible for redirecting the output and error streams, starting the process, and waiting for the process to end.

java/build/bazel/tests/integration/BazelCommand.java
ittaiz: Ok. I’ll live with it though I’m glad I don’t code in java in the d2d...

ittaiz:
ok, re feature envy I had a change ready which just removes the builder and has the existing code use the `Command`'s constructor but I decided to let it be since if we'll want an additional optional property which relates to process handling and not bazel commands in general we might want a builder back. Let's leave it.

One more thing- I think the toString of `BazelCommand` can become an important API for people troubleshooting their tests. Any chance for a followup PR with a simple test to cover us with respect to what we expect from the toString?

hchauvin:
You are right, I'll keep things as atomic as possible in the future :)

Issue #63, opened 2018-04-23
Share the download cache with the invoking Bazel installation


hchauvin:
... or have another way to share external dependencies.

Currently, the Bazel environment that is created by bazel-integration-testing has a download cache that is separated from the invoking Bazel installation (let's call it the "upstream" download cache).  It is possible to have a download cache when developing by pinning the TEST_TMPDIR, but since the download cache is a content-addressable storage, wouldn't it make sense to simply use the upstream download cache as it would be safe anyway?  In some cases this could greatly improve performance.  Agreed, such external dependencies should be avoided when doing integration testing, but sometimes it is more convenient.

This is not something that could be added right away, though, as it would require Bazel to publish in some way the location of its download cache.

ittaiz:
this is really interesting. @anchlovi is also facing some issues with setting up external dependencies for the `scratch workspace` and your solution sounds really interesting. @anchlovi wdyt?

ittaiz:
Where is `REPOSITORY_CACHE` that you used in the rules_kotlin PR defined?
Also it sounds strange that the `scratch_workspace` will have access to the shared cache directory since its out of its sandbox

ittaiz:
I'll rephrase the last sentence as a question. Were you able to run it with sandboxing?

hchauvin:
Yep, it runs with sandboxing, but because of a horrendous hack involving absolute paths.  Actually in local dev I hardcode this: -"Dbazel.repository.cache=/tmp/cache".  That's clearly not a portable solution :)

I thought of another way of sharing external dependencies, which might actually work across the spectrum, including remote execution, and without touching Bazel: "prewarming".

Here, bazel would be executed during the build phase to generate a download cache, so that the ~14s are only paid once; I actually like that better because you have less assumptions concerning how bazel work, and those assumptions might break when you have, e.g., remote execution.

ittaiz:
I don’t understand what you mean by executing bazel once in the build phase.
Can you elaborate?

ittaiz:
Oh interesting, so you’re thinking of somehow segregating the workspace
creation out so you can also run it as a genrule which will do “bazel
fetch”? But then it doesn’t have network access either

hchauvin:
Ah yes they are not supposed to access the network.  But does it mean that they can't?

[...]

PR #71, opened 2018-04-25
add a repository rule for download caching

hchauvin:
This repository rule allows to transparently declare external
dependencies in the scratch workspace, as one would do in the regular
workspace, but makes it so that no download actually occurs in the
scratch environment, allowing for better performance through caching
and better isolation (block-network can be used as a test tag even with
those external dependencies).  This repository rule is a no-op for bazel < 0.12.0.

See #63


java/build/bazel/tests/integration/WorkspaceDriver.java
```diff
@@ -26,14 +28,16 @@
...
 import java.util.Map;
 import java.util.Optional;
+import java.util.Properties;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
 
 public class WorkspaceDriver {
 
-  private static Path tmp;
+  static Path tmp;
```
ittaiz: why does this need to be public?

java/build/bazel/tests/integration/WorkspaceDriver.java
```diff
@@ -26,14 +28,16 @@
...
   private static Map<String, Path> bazelVersions;
   private static Path runfileDirectory = Paths.get(System.getenv("TEST_SRCDIR"));
+  static Properties properties;
```
ittaiz: why does this need to be public?
we have global mutable state here :(
can we make it private and expose a method to query specific properties?
maybe even make it non static to ease our own tests (not sure about this last bullet)

java/build/bazel/tests/integration/WorkspaceDriver.java
```diff
@@ -43,14 +47,43 @@
   private Path workspace = null;
 
   public static void setUpClass() throws IOException {
+    String configFile = System.getProperty("bazel.configuration");
```
ittaiz: wdyt about moving this to a private method?

java/build/bazel/tests/integration/WorkspaceDriver.java
```diff
@@ -43,14 +47,43 @@
...
+
+    // Create the distdir with the external dependency archives
+    String externalDeps = properties.getProperty("bazel.external.deps");
+    if (externalDeps != null && !externalDeps.isEmpty()) {
+      Path distdir = tmp.resolve("distdir");
+      Files.createDirectories(distdir);
+      try (Stream<Path> extDeps = Files.list(distdir)) {
+        extDeps.forEach(path -> {
+          try {
+            Files.delete(path);
```
ittaiz: won't this fail if the `path` represents a directory?

java/build/bazel/tests/integration/WorkspaceDriver.java
```diff
@@ -43,14 +47,43 @@
...
     }
+
     bazelVersions = new HashMap<>();
+
+    // Create the distdir with the external dependency archives
+    String externalDeps = properties.getProperty("bazel.external.deps");
+    if (externalDeps != null && !externalDeps.isEmpty()) {
+      Path distdir = tmp.resolve("distdir");
+      Files.createDirectories(distdir);
+      try (Stream<Path> extDeps = Files.list(distdir)) {
```
ittaiz: is this `quitelyCleanDirectory`? If so can we extract it to a private method?

java/build/bazel/tests/integration/WorkspaceDriver.java
```diff
@@ -269,14 +302,24 @@ public BazelCommand withEnvironmentVariable(String key, String value) {
     public Command build() {
         String bazelRcPath = bazelrcFile.map(p -> workspace.resolve(p).toString()).orElse("/dev/null");
 
+        Stream<String> experimentalDistdir;
```
ittaiz: isn't this a bit of an overkill?
not writing a lot of java these days but isn't a simple string preferable? Maybe an Optional<String> with a getOrElse of empty string?
Also can we move this also to a private method?

javatests/build/bazel/tests/integration/BUILD
```diff
@@ -31,8 +31,32 @@ bazel_java_integration_test(
     runtime_deps = [":BazelIntegrationTestNoSourcesSrc"],
 )
 
+genrule(
+    name = "test_archive",
+    outs = ["ext.zip"],
+    cmd = """
+# From https://github.com/bazelbuild/bazel/blob/88374614b62b17587453137814d264fdaf7bd70e/src/test/shell/bazel/external_integration_test.sh
```
ittaiz: how does this work if you can't download it on build time? I think it downloads it on build time and I'm not sure I understand why we need to

javatests/build/bazel/tests/integration/BazelBaseTestCaseTest.java
```diff
@@ -76,6 +76,22 @@ public void testJvmFlags() {
     org.hamcrest.MatcherAssert.assertThat(System.getProperty("foo.bar"), is("true"));
   }
 
+  @Test
+  public void testExternalDeps() throws Exception {
+    loadIntegrationTestRuleIntoWorkspace();
+
+    driver.scratchFile(
```
ittaiz: where is the assertion here?

java/build/bazel/tests/integration/WorkspaceDriver.java
hchauvin: Sorry, this package-private is a side effect of me working before the workspace was accessible in read-only.  I'll re-check everything anyway as I plan to rebase on the reformatted base.

java/build/bazel/tests/integration/WorkspaceDriver.java
hchauvin: So the reason I did that is that, since it's package-private, it's an implementation detail, not something that is exposed to the user.  I agree this is ugly, but my reasoning was that the best way to deal with that was, instead of having a bunch of static objects, to have a companion class/singleton where to put the config, but I thought this refactoring would be better done, in full, in another PR.

java/build/bazel/tests/integration/WorkspaceDriver.java
hchauvin: What do you mean?  Again I think that configuration should be encapsulated in a separate class anyway.

java/build/bazel/tests/integration/WorkspaceDriver.java
hchauvin: It should not really be possible, though, because that's not how distdir is set to work (it is a flat dir by design), and if you have a subdirectory, then something is very, very broken.

java/build/bazel/tests/integration/WorkspaceDriver.java
hchauvin: Ok.

java/build/bazel/tests/integration/WorkspaceDriver.java
hchauvin: This is completely an overkill, but I did not want to use a mutable list, which would be clearly indicated here, because I know I am dealing with Scala people here :)  getOrElse of empty string won't work, because you'll end up with an argument of length 0 passed to Bazel, it won't be happy.  I fear that anything else than mutability is an overkill here, but that's my opinion.

Ok for the private method.

javatests/build/bazel/tests/integration/BazelBaseTestCaseTest.java
hchauvin: Nowhere, this test is a no-op and should not be here, sorry for that.

javatests/build/bazel/tests/integration/BUILD
hchauvin: What do you mean?

hchauvin:
Before I go into the details, what do you think concerning the overall approach?

ittaiz:
@hchauvin I need you to press the "update branch" button here so that it will update the PR from master. I'll re-trigger the build and when it's green I'll merge.

damienmg:
For the approach I am fine with it but I would rather use the cache directly. The cache is actually documented but the documentation [was never imported](https://bazel-review.googlesource.com/c/bazel/+/6933). TBH I don't understand why they added distcache flag instead of simply documenting the cache format / making it nicer.

hchauvin:
So you think that, contrary to what is said behind the motivation of
distdir, the interface to the cache is actually likely to be stable? The
reason I used distdir is that the cache format is clearly marked as
internal details at least on the github master.

damienmg:
I cannot say about the future, I am no longer in the Bazel team, but the
cache structure hasn't changed since its creation. Which mean at the moment
it would work will all supported version of Bazel.

ittaiz:
> Awesome change!

I'm sorry, too many tabs... I meant the linting one. 

Will respond soon but in general I'm very excited about the approach

ittaiz:
@damienmg I don't think we can use the cache dir directly **inside** the scratch workspace because of sandboxing. @hchauvin's approach of using distdir circumvents that because he places the distdir inside the sandbox using runfiles.
Am I mistaken?

hchauvin:
@ittaiz I can actually use the cache instead of using the distdir, the only reason I used the distdir was because the cache was clearly marked as an implementation detail.  I just rebased and reformatted everything with the linter.

javatests/build/bazel/tests/integration/BUILD
ittaiz: 🤦‍♂️ missed that this is a comment...
The purpose of this genrule is an archive just for **our** own test? If so wdyt about renaming the output to something like `ext.zip`->`externalDependenciesArchiveForInternalTest.zip`? and maybe the name of the genrule to something else since it loads the name of the external repository and at least for me it wasn't clear how they interact

ittaiz:
@hchauvin thanks for the rebase! it makes reviewing much easier.
How can you use the cache itself? I think it won't work with remote execution since only declared inputs are submitted to the remote

hchauvin:
So I looked at what the repository cache actually looks like, @damienmg, and it's all in one class, and the architecture is really simple and I cannot see how it can change.  The reason I think experimental_distdir has been introduced is that repository_cache requires you to put each file to cache in a path that includes their digest, whereas experimental_distdir does not, and for some people it could add complexity to their builds.  But I don't see either why the repository_cache directory layout is not public API.

Therefore, as per your suggestion, I nuked experimental_distdir and replaced it with experimental_repository_cache.  Now, the trick works for all the Bazel versions supported by bazel-integration-testing.

@ittaiz It works the same way my version with distdir works.  All the files to download are actually declared as external_deps, then I rebuild the cache myself before invoking Bazel.

I had to nuke as well the test with "ext.zip" because it was not so easy to do with the repository_cache interface.  Instead, I make sure I "freeze" the repository cache by making it read-only, so that I'm sure Bazel won't try to download additional deps, and I have a test for that.  The reason I added this mechanism is that I found out that, somehow, Bazel has the ability to use the network anyway, even in sandboxing with block-network (I didn't investigate further).

Thank you both for the support on this patch!

tools/bazel_java_integration_test.bzl
```diff
@@ -97,15 +120,23 @@ def bazel_java_integration_test(
...
     if add_bazel_data_dependency:
-      data = data + [
-          "@build_bazel_bazel_%s//:bazel" % version.replace(".", "_")
-      ]
+      cur_data += ["@build_bazel_bazel_%s//:bazel" % version.replace(".", "_")]
     native.java_test(
         name = "%s/bazel%s" % (name, version),
-        jvm_flags = ["-Dbazel.version=" + version] + jvm_flags,
+        jvm_flags = [
+            "-Dbazel.configuration=$(location %s.properties)" % prop_rule
```
ittaiz: @damienmg do you remember if this needs to be $execpath or something? I have a recollection that $location is for build time and we need runtime

damienmg:
@hchauvin at the creation of the repository cache there was that idea of having it as an API but we weren't sure about the use case, so it never got far away. Seems like the Bazel team has gone another way now...

Seems like the freezing doesn't work on linux in Buildkite: java.nio.file.FileSystemException: /var/lib/buildkite-agent/.cache/bazel/_bazel_buildkite-agent/16fed208024f2a5f406c7a5382cf5f8b/bazel-sandbox/3936978874677246825/execroot/build_bazel_integration_testing/_tmp/cc02c3dac8b34566a4624971a114d7f6/cache/content_addressable/sha256/90a8e1603eeca48e7e879f3afbc9560715322985f39a274f6f6070b43f9d06fe/file: Read-only file system


Just an idea (maybe add TODO somewhere in the PR): it would be cool to be able to specify not only downloaded artifact but any artifact. E.g. I want //bin/my/special:artifact in the cache and just compute the checksum in the helper (so we can also use build artifact in the helper). I think of integration test where you have some constructed repository, constructed by bazel.

tools/import.bzl
```diff
@@ -0,0 +1,84 @@
...
+      "    ],",
+      "    visibility = [\"//visibility:public\"],",
+      ")",
+  ]
+
+  repository_ctx.file("BUILD", "\n".join(build_lines))
+
+bazel_external_dependency_archive = repository_rule(
+    attrs = {
+        "srcs": attr.string_list_dict(allow_empty = False),
```
ittaiz: Why is this a dict? AFAICT if I add two entries they will both be part of the same external repository in the scratch workspace right? Am I reading it wrong?

ittaiz:
Damien,
Why do you think the team went a different way?

Re your generalization- how do you see the scratch workspace utilizing the
“special:artifact”? The current example works because inside
java_import_external uses the downloaded who uses the cache

damienmg:
Ittai: I cannot comment, I am no longer in the team.

tools/bazel_java_integration_test.bzl
damienmg: @ittaiz: this needs to be the runtime location, not the execpath. IIRC location give the runtime location so it should be good.

tools/import.bzl
hchauvin: It's a dictionary of sha256 digests to download URLs.  The download cache works with digests, so the download URLs themselves are irrelevant in the scratch workspace, only the sha256 digests matter.

To clarify things a bit, the cache is only a directory with the following descendants:
content_addressable/sha256/<sha256 digest>/file

So what I do is I take the downloaded file for each external dependency, which I name as its sha256 digest, put it as a runfile and construct by hand a repository cache with the layout above.

This requires of course that the corresponding repository rule uses a sha256 digest that is explicitly given as a rule attribute.  That's why, in Java, maven_jar won't work, but java_import_external will.  (maven_jar should not be used anymore, IMO, because among other things the sha-1 checksum is optional and it is not possible to specify transitive dependencies, and I think it should be deprecated.)

I suggest you look at the two new test cases in WorkspaceDriverIntegrationTest to get a better perspective on all that.

hchauvin:
@ittaiz I don't follow you here.  Do you have an example in mind where it should not work?  As far as I know, all the repository rules for including external dependencies have to use some sort of rctx.download mechanism, and if they don't they should not hit the cache.

To be more precise, the class implementing the repository cache in Bazel is com.google.devtools.build.lib.bazel.repository.cache.RepositoryCache.  It is called only in two places: in com.google.devtools.build.lib.bazel.repository.MavenDownloader, where an sha-1 digest is used, and in com.google.devtools.build.lib.bazel.repository.downloader.HttpDownloader, with an sha256 digest.

The HttpDownloader case is I think handed properly, and the MavenDownloader cache is not supported (but only because sha-1 is used), but I think that maven_jar should be deprecated anyway.

If you really want to support maven_jar, adding support for sha-1 as well is easy, but I am not in favor of that.

tools/import.bzl
ittaiz: I looked at all of the code but I think that what I was missing (and maybe should be added as a comment) is the layout of the cache. I’m good with respect to this comment.  I think it only strengthens my other comment about having a test (maybe the existing ones) which has two entries in the dict

tools/bazel_java_integration_test.bzl
ittaiz: Sorry, I meant the runtime location. I think location gives the runtime location for native rules. Not for skylark.

ittaiz:
@damienmg- I meant what makes you say that it “seems the team went another way” and not “why do you think the team decided to go another way”? More importantly do you think the current implementation by @hchauvin will be a problem?
@hchauvin two examples are:
1. Git repository- doesn’t use rctx.download 
2. Damien’s example of an artifact of the repo
I’m much more interested in solving the first use-case though it’s not blocking for this PR

tools/bazel_java_integration_test.bzl
damienmg: From the [documentation](https://docs.bazel.build/versions/master/be/make-variables.html#location): _The expanded paths are relative to the runfiles directory of the \*\_test or \*\_binary rule._

So we are fine here as we are in a java_test.

(and also I meant *runfiles* not *runtime*)

ittaiz:
With respect to freezing I suggest to try and push a commit where you don’t call “freeze” and see if the freeze test fails of passes. If it passes there (meaning it’s frozen by Bazel itself) then maybe you can just check in Freeze if the Filesystem is read only or not and NOP if it is read only already

ittaiz:
From https://docs.oracle.com/javase/7/docs/api/java/nio/file/FileSystem.html:
A FileSystem can provide read-only or read-write access to the file system. Whether or not a file system provides read-only access is established when the FileSystem is created and **can be tested by invoking its isReadOnly method.** Attempts to write to file stores by means of an object associated with a read-only file system throws ReadOnlyFileSystemException.

damienmg:
I meant this new flag is experimental and doesn't tell us what will be its future. The other flag is no longer experimental and the format of the cache hasn't changed since its creation. One idea of the original design was actually to be able to zip the cache dir and bring it to machine without network access to use it.

Anyway the current state is using the cache structure works for all version of Bazel, using the distdir works only for latest release.

ittaiz:
ok, did some investigation over at #73 
if I remove the call to freeze the "freeze" test fails because the error is about not finding "repo1.maven.org" and not because of permission denied.
I removed the block-network tag to see if the test will pass that point and the bazel run will fail on downloading the jar since the filesystem is read-only but bazel succeeds.
I'm now trying to check if I remove the block-network but keep the `freeze` call how the test behaves.

ittaiz:
when calling freeze without block-network we still fail on RO filesystem.
@hchauvin any chance you can run the build of your branch on linux to see if this is bazel sandbox in general or something buildkite related?
I'm not sure why the filesystem is readonly with respect to permissions but readwrite when trying to add files

hchauvin:
@ittaiz Thank you for the work!  I've reach out for a Linux box with a configuration I use for buildkite agents, and the problem is due to the fact that sandboxing works differently on Linux than on Mac OSX (and is more robust on Linux): if you test with --spawn_strategy=standalone, the test passes just fine.  So that's sandboxing-related.

On Linux the block-network directive actually works, so what do you think about simply adding some comments to the freeze() method and silencing the IOException? BTW I now see the behavior of the sandbox on Mac OSX as a Bazel bug, but that shouldn't prevent us from patching together a PR.

hchauvin:
I integrated your suggestions in a new commit for better readability.  Feel free to fixup.

tools/bazel_java_integration_test.bzl
```diff
@@ -22,6 +22,27 @@ load(
     "bazel_binaries",
 )
 
+def _bazel_java_integration_test_properties_impl(ctx):
+  properties = [
+      "bazel.version=" + ctx.attr.bazel_version,
+      "bazel.workspace=" + ctx.workspace_name,
+      "bazel.external.deps=" +
+      ",".join([d.short_path for d in ctx.files.external_deps]),
```
ittaiz: @hchauvin this line collects all files of a filegroup from all filegroups? 
AFAICT each external dep is a bazel_external_dependency_archive which writes a filegroup with K entries.
Does the above collect all K entries from the V external dependencies?

java/build/bazel/tests/integration/RepositoryCache.java
```diff
@@ -0,0 +1,131 @@
...
+    this.cachePath = cachePath;
+  }
+
+  /** Add a file to the cache. The file name should be the sha256 digest of the file content. */
+  public void put(Path sourcePath) throws IOException {
+    Path sha256 = sourcePath.getFileName();
+    // TODO: assert the sha256 key is valid, maybe even compare to content?
+    Path cacheEntry = cachePath.resolve(CAS_DIR).resolve(KEY_TYPE).resolve(sha256);
+    Files.createDirectories(cacheEntry);
+    Files.createSymbolicLink(
```
ittaiz: @damienmg off-topic: wdyt about moving "copyFromRunfiles" to using a symlink?

javatests/build/bazel/tests/integration/WorkspaceDriverIntegrationTest.java
```diff
@@ -53,6 +53,91 @@ public void testRunningWithEnvironment() throws Exception {
...
+
+    Command cmd = driver.bazelCommand("test", "//:TestMe").build();
+
+    int returnCode = cmd.run();
+    assertEquals(0, returnCode);
+  }
+
+  /** Try (unsuccessfully) to download a dependency that was not specified as an external dep. */
+  @Test
+  public void testRepositoryCacheIsFrozen() throws Exception {
```
ittaiz: @hchauvin It looks to me like there's a bit of a mixup (though I might be wrong).
1. Why is the second URL of junit-4.11? I think this works but is confusing.
2. Why is jcommander specified here? Doesn't it belong in the happy path above to see that it can be used (the other entry from the dict)?

ittaiz:
@hchauvin Thanks! I did a small fixup and I want to add another test (or change the existing happy path) to show passing two different bazel_external_dependency_archive to a test.
I'm waiting with it since I'm not sure I understand the failure-case test (where I commented) and I'd rather we not have conflicts.

ittaiz:
@hchauvin I'd appreciate your review on my last two commits

ittaiz:
@hchauvin I’m thinking of a follow up to this API by exposing a method on WorkspaceDriver which given a name (file name?) returns the sha and URLs.
We have this info in skylark and it can remove yet another (third) repetition of the sha256 and URLs in the scratch workspace.
The con is that this is more logic to maintained and it’s a bit magical.
Wdyt?

PR #75, opened 2018-04-29
do bazel build/test on travis as well, add Mac support

hchauvin:


ittaiz:
Why?

hchauvin:
Sorry, I should have added a motivation.

1) Crowd thinking: other projects in the bazelbuild organization do that.  Some even support circleci.
2) The Buildkite bazelci runs on the Bazel team infrastructure, so any change to a PR must be approved before a build is launched there.
3) The travis-ci interface is public, everybody can look up the build history at any time.  bazelci only provides a crude interface for non-member (but apparently the Bazel team is working with Buildkite to improve that).
4) If somebody wants to fork the project and have CI on their fork, with travis-ci they only have to enable their project and no change has to be made on .travis.yml, with bazelci they must maintain their own CI infrastructure.

A concrete example: I have this available on my fork: https://travis-ci.org/hchauvin/bazel-integration-testing

ittaiz:
I don’t love this actually.
We suffered quite a lot from Travis over in rules scala and we’d like
actually to remove it when *this* library stabilizes :) (we have many e2e
outside of bazel)
@philwo I think this is due to the state of the CI (which is maturing...)
I think #2 and #3 need answers. At least the impact of #2 can be
significantly decreased IMO if we whitelist (manually?) contributors
Wdyt?

hchauvin:
I agree with your criticism of travis when it comes to Bazel.  I don't know about the problems you encountered with rules_scala,  but the infra I set up for private projects is based on the Buildkite CI script of the Bazel organization.  With this solution, you have the benefit of controlling your own infra with various sorts of caching and performance improvements, so that's a win.  I moreover agree 2 and 3 are shortcomings of the Buildkite interface and not fundamental flaws of bazelci.

4 (about forking) is IMO a more important point, and might or might not be something people factor in when they fork, but I can live without.  I'll rebase my branches before a PR to remove the travis yaml I use.

philwo:
Hi,

regarding 1), I think this is mostly for historical reasons. I don't see value in testing the same code on multiple CI services, unless the other service can provide you with e.g. a platform that you could otherwise not test on or a feature that you need (in both cases: let us know!).

Regarding 2), we're working on hardening our infrastructure and ideally would like to get rid of the "unblock" step for testing PRs. Also we have to revamp our permissions setup to make it easier to add people to the whitelist. Totally agree with that.

For 3), we're blocked on Buildkite implementing the anonymous access feature. They promised that it'll be ready "soon", but we don't have an exact date. Is there something you don't like about the current interface provided? There's a Google team maintaining and improving that and I'd be happy to relay feedback to them.

For 4), yes - we're not a service like Travis CI, where everybody can sign up. If you fork a project with the intent of maintaining a fork (and not just for PRs), then it's on you to setup supporting infrastructure. But if you fork a project in order to send a PR, this should already be fine, right? (Except for the annoying unblocking step.)

hchauvin:
Ok thanks for the feedback @philwo, I'm sold then :)

When and if https://github.com/bazelbuild/bazel/issues/4086 is solved, I would be able to run the linter with "bazel run" and so we could get rid of travis completely. 👍

ittaiz:
Will we be able to do “bazel run” on buildkite?

PR #77, opened 2018-04-30
support "bazel run" with arguments

hchauvin:
Bug: The repository cache was specified at the very end of the
arguments given to BazelCommand, meaning that with "bazel run //foo --
arg" it would end up as an additional argument to the executable.

### DIFF
--- a/java/build/bazel/tests/integration/WorkspaceDriver.java
+++ b/java/build/bazel/tests/integration/WorkspaceDriver.java
@@ -308,8 +308,18 @@ public class WorkspaceDriver {
                   "--max_idle_secs=10",
                   "--bazelrc=" + bazelRcPath));
 
-      command.addAll(args);
-      command.addAll(repositoryCache.bazelOptions());
+      // This would split the args "run //target -- hello world" into
+      // "run //target" and "-- hello world" ("hello world" being passed to the executable
+      // to run).
+      int terminator = args.indexOf("--");
+      if (terminator == -1) {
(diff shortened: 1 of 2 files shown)


java/build/bazel/tests/integration/WorkspaceDriver.java
```diff
@@ -308,8 +308,23 @@ public Command build() {
...
+      List<String> nonBazelArgs;
+      if (bazelArgsTerminator == -1) {
+        bazelArgs = args;
+        nonBazelArgs = Collections.emptyList();
+      } else {
+        bazelArgs = args.subList(0, bazelArgsTerminator);
+        nonBazelArgs = args.subList(bazelArgsTerminator, args.size());
+      }
+
+      command.addAll(bazelArgs);
```
damienmg: what about
```java
int terminator = args.indexOf("--");
if (terminator) {
  command.AddAll(args);
  command.AddAll(repositoryCache.bazelOptions());
} else {
  command.AddAll(args.subList(0, terminator));
  command.AddAll(repositoryCache.bazelOptions());
  command.AddAll(args.subList(terminator, args.size()));
}
```

hchauvin: I made room for options other than repositoryCache, but I like your solution better.

hchauvin:
If you are asking my opinion and not @damienmg: you can go with it, if there are conflicts, I'll rebase the other :)

Issue #64, opened 2018-04-23
Executable and runfiles "teleportation"


hchauvin:
Sorry in advance for the cheesy name :)

Imagine you want to write an integration test for a rule that depends on an executable with runfiles that is somewhat expensive to make (let's say, a rule for a language and you want to test some side effects of its compiler).  One way to do that with speed while retaining a certain level of isolation and generality would be to: 

1. Build the executable in the invoking bazel environment.
2. Gather all the runfiles in a runfile manifest.
3. Add the manifest, the executable and the executable's runfiles as runfiles to, e.g., bazel_java_integration_test.
4. Reproduce the runfile tree in the integration environment by symlinking with the information provided in 3.

Pseudo-code:

BUILD file:
```
java_binary(
    name = "bin",
    ...
)

teleportation_manifest(
    name = "bin_with_manifest",
    binary = ":bin",
)

bazel_java_integration_test(
    name = "test",
    data = [":bin_with_manifest"],
)
```

Test suite:
```
public class Test {
  ...

  @Before
  public void setUp() {
    driver.setUp();

    driver.teleportExecutable(
        WorkspaceDriver.getRunfile(".../bin_with_manifest.txt"),
        "to/some/package",
        "target");

    driver.scratchFile(
        "foo/BUILD",
        "sh_test(",
        "    ...,",
        "    data = [\"//to/some/package:target\"]",
        ")");
  }
}
```

hchauvin:
Oops I just saw #32 which is definitely relevant, sorry for the duplicate.  I also saw https://github.com/bazelbuild/rules_scala/pull/364.  To be precise, my suggestion does not imply modifying anything, it just adds a new feature that in my opinion address an important pain point.

I did a proof of concept with https://github.com/bazelbuild/rules_kotlin/pull/52 (about code coverage).  Relevant files are here: https://github.com/bazelbuild/rules_kotlin/tree/4c2b07476a9fe41334ed1e1f9dc96cfe9cdf4b23/tests/integrationtests/jvm/coverage

@ittaiz What do you have in mind when you allude to buildifier rules in #32?

======

Concerning a question that arose in https://github.com/bazelbuild/rules_scala/pull/364, one way I found to import the current workspace to the integration environment is to pass the workspace as data:

```
bazel_java_integration_test(
      ...,
      data = ["//:WORKSPACE"],
)
```

From that, you can unwrap the symlink and have an integration WORKSPACE as such:
```
local_repository(
[...]

PR #78, opened 2018-05-01
add an API to import an executable+runfiles

hchauvin:
This PR implements #64.  This is probably a work in progress concerning the API.

The feature here allows one to import to the scratch workspace an executable that was compiled outside of the scratch workspace.

The motivation is testing rules.  @ittaiz it might be of interest to you because, for instance, it allows you to import the scala compiler for testing the scala rules.  Having the download cache, this PR and copying all the *.bzl files is probably sufficient to have an integration test running.


ittaiz:
I’m actually unclear about the motivation since bazel works with source dependencies and IIUC this means the scratch workspace doesn’t depend on this executable via source but on it’s outputs.
Is this only for better performance?
If I can copy the bzl files do I *have* to do the above?

@damienmg you also mentioned this feature, what scenario did you have in mind?

ittaiz:
I re-read #64 and it indeed sounds like the main motivation is performance. I'm completely not belittling it but I want to understand if there are additional motivations?

hchauvin:
One point, of course, is performance, because if you can avoid recompiling a compiler for every one of your test cases, that could be great.  This is not something that can be accomplished in general by just copying the runfiles of the binary to the workspace.  With the JVM, it's a bit simpler because jars are easier to move around, and you could imagine a java_import of a fat (deploy) jar.

However, this proposal is also about convenience.  With it, you don't have to import all the dependencies that you declare in the WORKSPACE, your BUILD files or your source files.  You don't have to maintain a new way to build the executable just for your integration tests.  In my opinion, it simplifies the writing of integration tests a lot.

It focuses on executables with runfiles because when you implement a rule for a language in Bazel, you often have a modified compiler laying around (to have a Bazel worker, for code coverage, to resolve file names). In integration tests, your goal is to check how this compiler behaves with your Skylark rules, so wouldn't it make sense to have this executable already built?

Issue #34, opened 2018-03-25
Allow outputing logs of inner tests


ittaiz:
The idea is that we allow outputing the logs inner tests so that the developer can see in which way the inner test failed and maybe also assert on it.
Exists in logsOfInternalTests method https://github.com/bazelbuild/bazel-integration-testing/blob/master/javatests/build/bazel/tests/integration/BazelBaseTestCaseTest.java#L107

ittaiz:
cc @maximn

hchauvin:
Wouldn't it be better to use the Build Event Protocol to extract test outputs instead of attempting to parse stderr?  This is what bazelci with Buildkite is doing, as is the IntelliJ plugin.

One way to do that right now is to pass a temporary file to "--build_event_binary_file" then to parse it (binary protocol buffer), see https://github.com/bazelbuild/intellij/blob/7801ae72aa44d4057f4ff88f9774efc1c81d791e/base/src/com/google/idea/blaze/base/command/buildresult/BuildEventProtocolOutputReader.java

ittaiz:
Sounds like a better plan.
Would welcome a PR :)

PR #79, opened 2018-05-03
Collect the build event stream #34

hchauvin:


hchauvin:
One thing that probably should be discussed is how to import the build event stream dependency.  I shied away from including the Bazel code base as a dependency, and I borrowed a jar file from the IntelliJ plugin.

PR #81, opened 2018-06-15
moved to bazel version with jdk since without-jdk doesn't exist in ne…

ittaiz:
…wer releases


ittaiz:
So maybe we should cut back some of the older versions?

damienmg:
Yes let's do that.

ittaiz:
Can you please review?

ittaiz:
@damienmg fyi I'll merge tomorrow morning once the tests pass (only added a comment so confident they will pass again). I need to merge since this is blocking us internally but if you have any feedback about the copy of files in the repository cache I'd be happy to hear.

PR #83, opened 2018-08-14
Add CI config to test on RBE

ittaiz:
Don't know if this will work but I've just copied the change from rules_scala PR.
Let's see.

### DIFF
--- a/.bazelci/presubmit.yml
+++ b/.bazelci/presubmit.yml
@@ -27,4 +27,4 @@ platforms:
     test_flags:
     - "--test_output=all"
     test_targets:
-    - "..."
+    - "//... -//go/..."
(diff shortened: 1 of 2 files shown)


ittaiz:
@achew22 it seems RBE support is failing for go.
I can add RBE I think only to java part in the CI config but do you want to take a shot at supporting RBE? For us it was a substantial gain

achew22:
@ittaiz have you had a chance to look at this?

ittaiz:
Yes. It’s missing a handful of flags we need to add.
I expect this will be needed on a per driver level since this is what we’ve
added internally.
I hope next week to dedicate some time on this (before the summit)

PR #87, opened 2018-09-19
No rc in repo rule bazel version

### LINKED ISSUE #86: bazel_binary repository rule breaks if master rc is present on machine
Hi,
We build a docker image for our developers for several reasons the most significant one is maximizing cache hits. All of the bazel they run are inside the machine. 
To maximize cache hits they share many of the flags our CI machines use including platforms/toolchains and more.
We have these flags setup in a master rc file which is baked into the image.
When running the bazel_binary rule inside the container this fails since it picks up the master rc in the first "sanity" check of `bazel version`.
I'll send a PR shortly that disables all rc files only for this first call of the repository rule (`ignore_all_rc_files`).
cc @damienmg

ittaiz:
fixes #86 
cc @damienmg @or-shachar


ittaiz:
oh no. Our lint check is broken :(

damienmg:
https://github.com/bazelbuild/bazel-integration-testing/blob/master/private/format.bzl#L57 is pointing to a rather old version of Bazel, Travis use the latest Bazel. Since the java_toolchain used for Bazel is defined in the Bazel repository this is probably just it.

ittaiz:
maybe. In any case please don't merge yet since these flags don't work with older versions.
Trying to understand what works with older versions. We might need to drop some of the really old versions.

ittaiz:
ok
@damienmg I needed to go through quite a lot of yak shaving for this 3 word change :(
fixed linting script by updating versions and shas
fixed create hash dict script since 0.17.0 was never released and I needed to introduce the "concept" of skipped versions
would appreciate your review again

ittaiz:
ok, I'm about to throw in the towel :( there are tens of issues due to the upgrade

ittaiz:
already with bazel 13 and buildifier 12 there are many issues 
I'm thinking of disabling the check and opening an issue for us to re-instate it wdyt?

tools/bazel_hash_dict.bzl
```diff
@@ -1,28 +1,20 @@
...
-    "0.14.1": {
-        "darwin-x86_64": "0bdf63909ef9ba446a2d7cd077edb3aaf4d63b725e82e7cdcb69df482df61a3c",
-        "linux-x86_64": "7b14e4fc76bf85c4abf805833e99f560f124a3b96d56e0712c693e94e19d1376",
-    },
     "0.15.2": {
         "darwin-x86_64": "84ee271274e7c8904a48f7894fe0569174d162c7428a19dda57fc6a6c95c628b",
         "linux-x86_64": "13eae0f09565cf17fc1c9ce1053b9eac14c11e726a2215a79ebaf5bdbf435241",
     },
+    "0.16.0": {
+        "darwin-x86_64": "bb4720c027a991643be7dabee5c96a84c3776d6e2da92bf65df012a7cec15569",
```
damienmg: Maybe not have 0.16.0 since there is already 0.16.1?

ittaiz:
Thanks!
Not sure I understand what you mean by fixing the test? I already ran
“./lint.sh fix” locally and pushed. Is this what you meant?
Also, re 0.16.0 this is the current logic of the create dict. I can
manually edit it.
Wdyt?

damienmg:
I meant that normally running buildifier should have fix it but if not then
there is more to it.

Yes I meant manually edit it. I don't know why the update script added
0.16.0, it correctly passed 0.17.0...

ittaiz:
0.16.0 exists.
0.17.0 doesn’t exist and i needed to update the script to ignore 0.17.0
(and get to 0.17.1).
Sounds good.
I’ll merge it as is and open an issue to revive the linting.
Honestly we use this library a lot internally but our biggest pains
currently are the fact that it can’t run on RBE and the infinite symlink
expansion.
These are serious issues we want to resolve.

ittaiz:
please don't delete the branch just yet. We need it since we currently point to it and not to master (we'll change next week)

PR #91, opened 2018-09-27
Remove pre-extract of bazel binary

### LINKED ISSUE #88: infinite symlink expansion
Easy to reproduce after running build 2-3 times.
```
ERROR: infinite symlink expansion detected
[start of symlink chain]
some-path/external/build_bazel_bazel_0_16_1/install_base/_embedded_binaries/_embedded_binaries
some-path/external/build_bazel_bazel_0_16_1/install_base/_embedded_binaries
[end of symlink chain]
ERROR: some-other-path/bazel-tooling/external_third_parties/src/it/BUILD.bazel:26:1: no such package '@build_bazel_bazel_0_16_1//': error globbing [install_base/**]: Infinite symlink expansion and referenced by '//external_third_parties/src/it:third_parties_test_runner/bazel0.16.1'
ERROR: Analysis of target '//external_third_parties/src/it:third_parties_test_runner/bazel0.16.1' failed; build aborted: no such package '@build_bazel_bazel_0_16_1//': error globbing [install_base/**]: Infinite symlink expansion
```
cc @damienmg any quick tips pop into your mind on how to solve this?

### LINKED ISSUE #90: bazel integration testing doesn't work with RBE
Due to the extraction optimization we do in the repository rule Bazel can't run on RBE.
I've worked around many of the parts by manipulating the mtime but there are more files needed (for example `$INSTALL_BASE/_embedded_binaries/embedded_tools/platforms/BUILD`).
I've experimented with using @damienmg's original unpacked version and it works on RBE.
We pay a cost of re-extracting bazel every test run which is currently ~10 seconds.
This is significant locally but trivial on CI given we can now parallelize (when one uses RBE the build server agents themselves are often weak and can't run many concurrent e2es) and cache hits.
Given the [work](https://docs.google.com/document/d/1Igmv-2GfXkoVFWTXvBYPeniQom8nLAwzqzridDlBIS4/edit?usp=sharing) in bazel on decreasing the binary size and decreasing extraction time it's reasonable to assume sometime in the next 6 months this cost will drop to 3-5 seconds.

@damienmg I want to remove the pre-extraction and just use the original bazel.
WDYT?

ittaiz:
fixes #90 
pre-extraction causes many issues with transmitting the extracted bazel files to rbe (preserving timestamps, various files which aren't copied for some reason).
I think this might also fix #88 since I haven't seen it since running with this mode.  
The failures are linting related :(


ittaiz:
Wdyt about merging this and trying to iterate on the TreeArtifacts solution?
Also,
I think part of the problem is that Bazel fails if mtime of any of the
files is not in the future (years away) BUT RBE doesn’t preserve mtime
since that would make cache hits irrelevant.
We can stamp the mtime in the invocation but I’m not sure there aren’t
other caveats.
I want the pre extract benefit but I’ve spent many many hours iterating on
this issue and this solution at least works now.

damienmg:
I am fine with merging this one (that's what I meant earlier). I don't even
know if we need to do the treeartifact solution. Wix were the one hitting
the slowness iirc so it is up to you, if it's not a big pain it might not
be worth doing more.

PR #92, opened 2018-10-01
Add GET_LATEST_BAZEL_VERSIONS method

achew22:
It returns a certain number of bazel versions. It allows you to test the
latest 3 minor versions of Bazel without testing all the patch versions
or having to hardcode any values.

### DIFF
--- a/tools/common.bzl
+++ b/tools/common.bzl
@@ -16,13 +16,13 @@ load("//tools:bazel_hash_dict.bzl", "BAZEL_HASH_DICT")
 
 BAZEL_VERSIONS = BAZEL_HASH_DICT.keys()
 
-def zfill(v, l=5):
+def _zfill(v, l=5):
   """zfill a string by padding 0s to the left of the string till it is the link
   specified by l.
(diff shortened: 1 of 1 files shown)


tools/common.bzl
```diff
@@ -15,3 +15,51 @@
 load("//tools:bazel_hash_dict.bzl", "BAZEL_HASH_DICT")
 
 BAZEL_VERSIONS = BAZEL_HASH_DICT.keys()
+
+def zfill(v, l=5):
```
ittaiz: Can we make this private?

tools/common.bzl
```diff
@@ -15,3 +15,51 @@
...
+  """zfill a string by padding 0s to the left of the string till it is the link
+  specified by l.
+  """
+  return "0" * (l - len(v)) + v
+
+def unfill(v, l=5):
```
ittaiz: Can we also make this private?

achew22:
Thanks for the review. PTAL

PR #99, opened 2018-12-05
Update rules_go to 0.16.3

achew22:
Fixes the build


achew22:
FYI, you can set up renovatebot (which handles WORKSPACE and BUILD files) on the repo. If it is just a version bump it handles it very well. If it is more complex and requires changing the WORKSPACE/BUILD then this will need to happen. But at least it means you're only doing the interesting upgrades.

ittaiz:
Thanks! Is there some documentation for renovatebot?

achew22:
https://renovatebot.com

PR #100, opened 2019-01-15
Fix common.bzl

meteorcloudy:
To fix incompatible change: [incompatible_disallow_slash_operator](https://github.com/bazelbuild/bazel/issues/5823)

Fixed by `buildifier --lint=fix ./tools/common.bzl`


meteorcloudy:
@achew22 Can you deal with the travis failures?

ittaiz:
Travis uses an old version of linting which is broken :(
Need to fix it.
Merging as is. Thanks!

Issue #85, opened 2018-08-15
[Bug] bazel_java_integration_test doesn't pass tags to suite


ittaiz:
If one sets tags on `bazel_java_integration_test` they are propagated to the individual targets but not set on the suite.
We saw this as a problem when we set the `manual` tag on the target but they still run (because the individual tests are depended upon by the suite which gets called in the label expansion).
It's basically here but the test will take me a bit and I'm in a different context.  
https://github.com/bazelbuild/bazel-integration-testing/blob/master/tools/bazel_java_integration_test.bzl#L144

PR #103, opened 2019-01-23
Add CI config to test on RBE

ittaiz:
superseeds #83

### DIFF
--- a/.bazelci/presubmit.yml
+++ b/.bazelci/presubmit.yml
@@ -2,22 +2,31 @@
 platforms:
   ubuntu1604:
     build_targets:
-    - "..."
+    - "//..."
     test_targets:
-    - "..."
+      - "--"
+      - "//..."
+      - "-//javatests/build/bazel/tests/integration:RBEExampleTest/bazel0.22.0"
+      - "-//javatests/build/bazel/tests/integration:RBEExampleTest/bazel0.23.2"
   ubuntu1804:
     build_targets:
-    - "..."
+    - "//..."
     test_targets:
-    - "..."
+      - "--"
+      - "//..."
+      - "-//javatests/build/bazel/tests/integration:RBEExampleTest/bazel0.22.0"
+      - "-//javatests/build/bazel/tests/integration:RBEExampleTest/bazel0.23.2"
   macos:
     build_targets:
-    - "..."
+    - "//..."
     test_targets:
-    - "..."
+      - "--"
+      - "//..."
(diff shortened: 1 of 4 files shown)


achew22:
It seems to me, now that BuildKite builds from any contributer, that we should consolidate on a single CI system. WDYT about removing the Travis tests? They seem to be an order of magnitude slower, and much flakier.

achew22:
@ittaiz have you had a chance to look at this?

ittaiz:
Yes. There is a set of flags that are needed for every test to pass so it
will work.
I need to see how to expose them so the tests can use them

achew22:
Are these flags to Bazel itself, flags to the test, or flags to the RBE?

ittaiz:
Flags the test code needs to pass to the “embedded” bazel

ittaiz:
@achew22 this is ready for review.

Pro: 
1.  Finally enabling RBE on CI of this library.
2. Users can have an example of what's needed to run a bazel integration test on RBE.

Con:
1.  Only a single test.
2. Solution isn't baked into the library yet.

All in all I think this is an important step forward.
cc @nlopezgi

.bazelci/presubmit.yml
```diff
@@ -2,22 +2,31 @@
 platforms:
   ubuntu1604:
     build_targets:
-    - "..."
+    - "//..."
     test_targets:
-    - "..."
+      - "--"
+      - "//..."
+      - "-//javatests/build/bazel/tests/integration:RBEExampleTest/bazel0.22.0"
```
achew22: Could you leave a comment explaining why this version is excluded. I assume it's because RBE wasn't complete till 0.24.x, but am not sure.

ittaiz: Not sure I understand what you mean.
This version isn't excluded but rather all instances of `RBEExampleTest` are excluded. It has two instances (22/23) because of toolchains issue.
It's excluded because this test can only be run on RBE.
The real solution is to bake RBE support into the library (for example internally we use a define to say if it's using RE or not).
A better solution would have been to tag this test as something and excluded these kind of tests but due to #85 that isn't an option right now.
Does that answer the question?

ittaiz: I'll merge this as is and I'll just try to fix #85 sometime this week. 
Actual fix is really small, test will take a bit more.
Once that's done I'll tag the RBE test and it will be clearer.

PR #105, opened 2019-02-18
Add recent Bazel releases

achew22:


ittaiz:
@achew22 did you use the tool to create this? Asking since I think we configured the tool to retain less versions. Each version has its cost on CI as well as needing to be backward compatible for a long time

achew22:
I did run the tool but I'm not attached to the PR. If you wanna regenerate with your own parameters that would be great

ittaiz:
done as part of windows support

PR #112, opened 2019-02-27
Reorganize the codebase and

achew22:
This refactors the java tests


javatests/build/bazel/tests/integration/BazelBaseTestCaseTest.java
```diff
@@ -49,28 +47,26 @@ public void testRunningBazelInGivenRelativeDirectory() throws Exception {
...
-    org.hamcrest.MatcherAssert.assertThat(System.getProperty("foo.bar"), is("true"));
-  }
-
-  private void setupPassingTest(final String testName) throws IOException {
-    writePassingTestJavaSource(testName);
-    writeTestBuildFile(testName);
-  }
-
-  private void writeTestBuildFile(final String testName) throws IOException {
+    final String testName = "IntegrationTestSuiteTest";
```
ittaiz: I have to be honest and say I feel these were not indirection but abstraction.
The current state feels, to me, as an overload of details and literals.
I'd really like for us to keep the previous state.

ittaiz:
Also the build doesn't pass (only on mac which is strange)

ittaiz:
@acew22 did my review discourage you?

PR #96, opened 2018-11-11
Add Windows support

jchv:
WIP. Depends on the update to `rules_go` in #95.

Java tests are failing quietly, I can't find any relevant logs yet.

PR #117, opened 2019-03-12
Add Windows Support

meteorcloudy:
Inspired by https://github.com/bazelbuild/bazel-integration-testing/pull/96, I also gave it a try.
I enabled all tests but `BazelBaseTestCaseTest` because it tries to run Bazel from a directory path longer than 260 characters, which caused the following error:
```
Executing tests from //:IntegrationTestSuiteTest/bazel0.23.2
-----------------------------------------------------------------------------
Error: Current working directory has a path longer than allowed for a
Win32 working directory.
Can't start native Windows application from here.

external/bazel_tools/tools/test/test-setup.sh: line 232: C:/src/tmp/l7b3czla/execroot/build_bazel_integration_testing/_tmp/b8fc8754c3be87e0fe08dc8ef9e1fd51/ajlaysar/execroot/build_bazel_integration_testing/bazel-out/x64_windows-fastbuild/bin/IntegrationTestSuiteTest/bazel0.23.2.exe: File name too long
/c/src/tmp/l7b3czla/execroot/build_bazel_integration_testing/_tmp/b8fc8754c3be87e0fe08dc8ef9e1fd51/ajlaysar/execroot/build_bazel_integration_testing/bazel-out/x64_windows-fastbuild/testlogs/IntegrationTestSuiteTest/bazel0.23.2/test.log (END)
```
@laszlocsomor Does the native test wrapper supports running Bazel from a long path? Maybe it could help fix this issue.

Another test case `testRepositoryCacheIsFrozen` in `WorkspaceDriverIntegrationTest.java` depends on the `WorkspaceDriver.freeze()`. The `freeze()` function doesn't work on Windows, and there is basically no way to prevent other process writing to a directory. So I cannot find any other way to freeze the cache. I worked around this by changing the url to a non-existing host which caused a "Unknown host" error to be caught. @ittaiz @hchauvin Is this ok with you?

### DIFF
--- a/java/build/bazel/tests/integration/WorkspaceDriver.java
+++ b/java/build/bazel/tests/integration/WorkspaceDriver.java
@@ -70,7 +70,10 @@ public class WorkspaceDriver {
   }
 
   private static void setupTmp() throws IOException {
-    String environmentTempDirectory = System.getenv("TEST_TMPDIR");
+    // We have to use a shorted output user root on Windows, otherwise we get
+    // a "current working directory is too long" error, so use TMP instead of "TEST_TMPDIR"
+    String environmentTempDirectory =
+      OS.getCurrent() == OS.WINDOWS ? System.getenv("TMP") : System.getenv("TEST_TMPDIR");
     if (environmentTempDirectory == null) {
       tmp = Files.createTempDirectory("e4b-tests");
       tmp.toFile().deleteOnExit();
--- a/java/build/bazel/tests/integration/WorkspaceDriver.java
+++ b/java/build/bazel/tests/integration/WorkspaceDriver.java
@@ -95,7 +95,8 @@ public class WorkspaceDriver {
   private static void unpackBazel(String version) throws IOException, InterruptedException {
     if (!bazelVersions.containsKey(version)) {
       // Get bazel location
-      Path bazelFile = runfile("build_bazel_bazel_" + version.replace('.', '_') + "/bazel");
+      String bazelName = "bazel" + (OS.getCurrent() == OS.WINDOWS ? ".exe" : "");
+      Path bazelFile = runfile("build_bazel_bazel_" + version.replace('.', '_') + "/" + bazelName);
       if (!Files.exists(bazelFile)) {
         throw new BazelWorkspaceDriverException("Bazel version " + version + " not found");
       }
@@ -178,14 +179,17 @@ public class WorkspaceDriver {
       throw new BazelWorkspaceDriverException(
           "The `stripPrefix` MUST be a prefix of `directoryToCopy`");
 
-    Path stripPrefixPath = Paths.get(stripPrefix);
+    // Paths.get("").relativize(Paths.get("foo/bar")) returns ..\foo\bar on Windows,
+    // but it returns foo/bar on Linux.
+    // Adding ./ as a prefix to all path to make sure relativize returns correct result.
+    Path stripPrefixPath = Paths.get("./" + stripPrefix);
     try (Stream<Path> paths = Files.walk(startingDirectory)) {
       paths
           .filter(path -> Files.isRegularFile(path))
           .forEach(
               runfilePath -> {
                 Path relativeToRunfilesPath = runfileDirectory.relativize(runfilePath);
-                Path destinationPath = stripPrefixPath.relativize(relativeToRunfilesPath);
+                Path destinationPath = stripPrefixPath.relativize(Paths.get("./").resolve(relativeToRunfilesPath));
                 try {
                   copyFromRunfiles(relativeToRunfilesPath.toString(), destinationPath.toString());
                 } catch (IOException e) {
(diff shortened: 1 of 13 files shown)


laszlocsomor:
> @laszlocsomor Does the native test wrapper supports running Bazel from a long path? Maybe it could help fix this issue.

No. CreateProcessW fails with 208 if the program's path in `lpCommandLine` is long, or with 267 if the `lpCurrentDirectory` is long.

meteorcloudy:
@ittaiz @hchauvin This change is ready for review, the tests are passing on Windows.

One question, do we actually need the `freeze` function in `WorkspaceDriver`? Because I cannot find any way to implement that on Windows.

java/build/bazel/tests/integration/WorkspaceDriver.java
```diff
@@ -66,7 +66,7 @@ private static void setupRepositoryCache() throws IOException {
         repositoryCache.put(Paths.get(dep));
       }
     }
-    repositoryCache.freeze();
+    // repositoryCache.freeze();
```
achew22: Rm?

javatests/build/bazel/tests/integration/BUILD
```diff
@@ -11,6 +11,9 @@ bazel_java_integration_test(
         "//java/build/bazel/tests/integration",
         "@com_google_truth//jar",
     ],
+    # We had to disable this test on Windows because it will fail due to running Bazel
+    # from a path longer than 256 charactors.
+    tags = ["no_windows"],
```
achew22: Can you file an issue for this?

javatests/build/bazel/tests/integration/BUILD
```diff
@@ -40,8 +43,5 @@ bazel_java_integration_test(
         "@test_archive",
         "@test_archive2",
     ],
-    # Ensure that there could be no network fetch in sandboxing.  This tag is a Bazel
-    # built-in, it is important to keep it.
-    tags = ["block-network"],
```
achew22: Why did you need to remove this tag? Does this make it not hermetic?

javatests/build/bazel/tests/integration/BUILD
meteorcloudy: Better than that, I figured out a workaround ;)

java/build/bazel/tests/integration/WorkspaceDriver.java
meteorcloudy: I'm waiting @ittaiz 's comment on this. If `repositoryCache.freeze()` is only for testing `testRepositoryCacheIsFrozen`, I think we can just remove it. Because I changed the url in `testRepositoryCacheIsFrozen` to a unknown host, which would produce the expected error on all platform.

javatests/build/bazel/tests/integration/BUILD
meteorcloudy: Because with newer version of Bazel, `WorkspaceDriverIntegrationTest` will fail with "unknown host" when it tries to access "mirror.bazel.build", I'm not sure how to deal with this without removing this "block-network" tag. @ittaiz Can you give some instructions?

achew22:
To me this is good to merge. I'm going to give @ittaiz 24 hours (March 15th at 1pm PDT, 4pm EDT) then I'll submit this. After that I'll update the iBazel binary to the latest version and push out a `0.10` release that theoretically works with windows :smile: 

GREAT work!

achew22:
Turns out I have a bunch of stuff going on that would prevent me from doing this until tomorrow afternoon unless I do it now and we are pretty close to the deadline I prescribed. If there is an issue with this then I think we should do a follow up pr. Thanks so much for sending this in! Merging

ittaiz:
Finally taking a look. I think we will have a few issues. I'll mention them here even though it's merged

ittaiz:
Update- we have one main issue with the repositoryCache.freeze and the test.

ittaiz:
another update- this PR introduced a significant issue in the form of downloading remote_jdk.
Since 0.21 Bazel will try by default to download a remote jdk unless instructed otherwise. When @meteorcloudy changed (and effectively removed) the frozen repository cache test they actually removed the one test that verified this library works with network isolation.

This issue happened because this PR did a couple of things since a PR that only updated the versions surfaced this issue [PR](https://github.com/bazelbuild/bazel-integration-testing/pull/105), [log](https://buildkite.com/bazel/bazel-integration-testing/builds/362#57e26e8a-4dfe-4978-83eb-717a41f91849).

I'm working on a solution but it's not baked in yet.
@cushon @lberki,
If I need to run bazel in a test with network isolation would you recommend passing `JavaRuntimeInfo.java_home` and then using `--host_javabase=@bazel_tools//tools/jdk:absolute_javabase` with a `--define=ABSOLUTE_JAVABASE=` which will point to the path I pass from the outside (dir in runfiles)?
Is there a simpler solution?
Thanks!

meteorcloudy:
@ittaiz Thanks for looking into the issues, sorry I didn't consider it in advance, please tell me if I can help with anything!

ittaiz:
Sure.
I think the main takeaway is to be super careful when changing existing
tests.
I'm waiting for a bit of input from Irina/Cushon about my take (
https://groups.google.com/forum/#!topic/bazel-sig-jvm/JrJldaI40PM) and
hopefully I will be able to continue

Issue #106, opened 2019-02-18
Java runner isn't compatible with Bazel > 0.21


achew22:
Sample run output: https://storage.googleapis.com/bazel-untrusted-buildkite-artifacts/57e26e8a-4dfe-4978-83eb-717a41f91849/javatests/build/bazel/tests/integration/BazelBaseTestCaseTest/bazel0.21.0/attempt_1.log

```
STDERR:
    $TEST_TMPDIR defined: output root default is '/home/bazel/.cache/bazel/_bazel_bazel/ec321eb2cc2d0f8f91b676b6d4c66c29/sandbox/linux-sandbox/57/execroot/build_bazel_integration_testing/_tmp/5779ebb4e6ed1d2af99650167aeb7cfb' and max_idle_secs default is '15'.
    Starting local Bazel server and connecting to it...
    INFO: Invocation ID: 417d9c78-6445-403c-8056-d4b8deee9464
    Loading: 
    Loading: 0 packages loaded
    Loading: 0 packages loaded
    Analyzing: 13 targets (1 packages loaded)
    Analyzing: 13 targets (1 packages loaded, 0 targets configured)
    Analyzing: 13 targets (3 packages loaded, 16 targets configured)
    Analyzing: 13 targets (18 packages loaded, 353 targets configured)
    Analyzing: 13 targets (18 packages loaded, 353 targets configured)
    ERROR: /home/bazel/.cache/bazel/_bazel_bazel/ec321eb2cc2d0f8f91b676b6d4c66c29/sandbox/linux-sandbox/57/execroot/build_bazel_integration_testing/_tmp/5779ebb4e6ed1d2af99650167aeb7cfb/0e3c39548b556d9f0eec702c5dd0dd08/external/bazel_tools/tools/jdk/BUILD:305:1: no such package '@remotejdk_linux//': com.google.devtools.build.lib.vfs.FileAccessException: /home/bazel/.cache/bazel/_bazel_bazel/ec321eb2cc2d0f8f91b676b6d4c66c29/sandbox/linux-sandbox/57/execroot/build_bazel_integration_testing/_tmp/5779ebb4e6ed1d2af99650167aeb7cfb/cache/content_addressable/sha256/f27cb933de4f9e7fe9a703486cf44c84bc8e9f138be0c270c9e5716a32367e87 (Permission denied) and referenced by '@bazel_tools//tools/jdk:remote_jdk'
[...]

PR #120, opened 2019-04-01
Revive freezing

ittaiz:
This solves #106 and the regression introduced in #117 
@lberki @iirina @cushon 
I'd really love your feedback about this change (in the bzl file and in WorkspaceDriver.java).
I'm mainly unclear why these are the flags and their values.
I expected the following to be needed:
```
        "--host_javabase=@bazel_tools//tools/jdk:absolute_javabase",
        "--javabase=@bazel_tools//tools/jdk:absolute_javabase",
        "--java_toolchain=" + toolchain
        "--host_java_toolchain="  + toolchain,
        "--define=ABSOLUTE_JAVABASE="+ javaHome
```
But instead the following was sufficient (which I don't understand):
```
 "--host_javabase=@bazel_tools//tools/jdk:jdk",
 "--java_toolchain=" + javaToolchain
```
FYI @achew22


iirina:
@ittaiz can you briefly describe what issue you're trying to solve and why do you expect those flag to be necessary?

ittaiz:
Yes, definitely.
What we're trying to do here (as shown in the failure of the first commit) is to run bazel itself inside a sandboxed network-blocked bazel test.
The problem is that bazel 21 and onwards tries to download a remote jdk if you don't explicitly specify it via ` --host_javabase` flag. This is a problem since I don't want the test to download the JDK.

What I'd like to do is have the starlark rule pass the location of the java runtime and have the test bazel use that for it's compilation.

ittaiz:
I mainly don't understand how `@bazel_tools//tools/jdk:jdk` helps (and will it be supported in the future) and why I need to pass  `--java_toolchain` but not `--host_java_toolchain`.
If this is by incident then I'd love to know what are the use-cases that require passing `host_java_toolchain` so I can write the relevant test for it.

ittaiz:
Thanks.
I’m not 100% sure about this and already found a bug case (bazel query
doesn’t accept host_javabase).
I’ll fix that and hopefully get some help from Irina/Liam

iirina:
If my understanding is correct, you need to use `@bazel_tools//tools/jdk:absolute_javabase` instead of `@bazel_tools//tools/jdk:jdk` only if your local JDK is at a location that couldn't be identified by bazel (e.g. by looking at `JAVA_HOME`) or if you want to use a different JDK if multiple are installed. Are any of these an use case in this test? In your initial comment how was `javaHome` from `"--define=ABSOLUTE_JAVABASE="+ javaHome` computed?

Setting `--host_javabase` is enough for not downloading the remote jdk. I assume you need to set `--java_toolchain` for being able to use a certain language level (or `jvm_opts`). I don't see a reason why `--host_java_toolchain` needs to be set as well. The default host java toolchain uses a javac embeded into bazel [*] which works with the default  `jvm_opts`.

You may want to use a different `--host_java_toolchain` if you have host java tools that need to be compiled with a different javac version than the one embeded into bazel. I can't think of another use case.

I hope this clears things a bit. 

[*] the javac (modules) may be embeded into bazel or into the remote java tools depending on the bazel version you're using.

ittaiz:
> Are any of these an use case in this test
I think it can be relevant if someone uses this on a machine that doesn't have `JAVA_HOME`. I'll reproduce this locally and then add a comment, thanks! (Maybe it should be documented in Bazel also somewhere? Maybe in the build file?)
The `javaHome` was computed from `ctx.attr._java_runtime[java_common.JavaRuntimeInfo].java_home_runfiles_path` where `_java_runtime` is `@bazel_tools//tools/jdk:current_java_runtime`.
I think I need to pass in specific `jvm_opts` of 8 because of the change here https://github.com/bazelbuild/bazel/commit/137504e8cacda23efcc4b74f7c1b03e0074da030.
Actually I'll try to see if I can refrain from passing this flag if the jdk is >8.

iirina:
The following part looks like a cycle to me:

```
"--javabase=@bazel_tools//tools/jdk:absolute_javabase",
"--define=ABSOLUTE_JAVABASE="+ javaHome",
```

Because `javaHome` is looking at `@bazel_tools//tools/jdk:current_java_runtime` which actually points to the value of `--javabase`. This is an interesting situation, I have to investigate how we handle this kind of cycle.

iirina:
In your test setup `@bazel_tools//tools/jdk:current_java_runtime` is evaluated first to the value of either JAVA_HOME or the `--javabase` value that was passed to the test.

When actually running bazel inside the test ABSOLUTE_JAVABASE points to the local JDK or the test's `--javabase` (if it was specified). Are you running the test with a custom javabase? If not, then

```
"--javabase=@bazel_tools//tools/jdk:absolute_javabase",
"--define=ABSOLUTE_JAVABASE="+ javaHome",
```

is equivalent to `--javabase=@bazel_tools//tools/jdk:jdk`, which points to the local JDK, so the default value in bazel.

ittaiz:
Thanks for your help!
If my “external” bazel is run with a javabase but the “test” bazel runs
without then I need the absolute approach right?

iirina:
> If my “external” bazel is run with a javabase but the “test” bazel runs without then I need the absolute approach right?

That is correct.

tools/bazel_java_integration_test.bzl
```diff
@@ -36,6 +38,9 @@ bazel_java_integration_test_properties_ = rule(
     attrs = {
         "bazel_version": attr.string(mandatory = True),
         "external_deps": attr.label_list(allow_files = True),
+        "_java_runtime": attr.label(
```
iirina: You can rename `_java_runtime` to `_test_java_runtime` or something similar to distinguish between the test external javabase and the internal one...

ittaiz:
Thanks!
Not sure that would be correct. I agree that’s whAt I’m using it for but
OTOH I’m not using a custom java runtime for the test but rather passing
the real java runtime into the test.

ittaiz:
@achew22 Major changes since you reviewed:
1. I'm using absolute javabase + define to support cases where people have multiple JDKs installed.
2. Supporting opt out of these flags per bazel invocation to support running bazel query.
3. Removed versions below 0.21 since they don't have absolute javabase

Note: I got this to work on our many internal tests *locally* but still haven't gotten the new flavor to work with RBE. Might mean more changes to this PR though we can also merge this and iterate further in a new PR.
Would appreciate your review

ittaiz:
Update: I got our tests to pass also on RBE :)
This is good to go

ittaiz:
@achew22 since this is very close to the original PR I'll merge it on Monday morning Israel time. Would still appreciate your second look if you get the chance before then.

java/build/bazel/tests/integration/WorkspaceDriver.java
```diff
@@ -66,7 +72,9 @@ private static void setupRepositoryCache() throws IOException {
         repositoryCache.put(Paths.get(dep));
       }
     }
-    // repositoryCache.freeze();
+    //freeze doesn't work on Windows
+    if (OS.getCurrent() != OS.WINDOWS)
+      repositoryCache.freeze();
```
achew22: Should we indicate to the caller that freezing is not supported in some way?

java/build/bazel/tests/integration/WorkspaceDriver.java
```diff
@@ -261,13 +270,77 @@ private BazelWorkspaceDriverException(String message) {
...
+  private BazelCommand.Builder bazel(List<String> args, boolean addJavaBaseConfigFlags) {
+    return new BazelCommand.Builder(this, tmp, repositoryCache,
+        concat(args, maybeJavaBaseConfigFlags(addJavaBaseConfigFlags)));
+  }
+
+  private List<String> concat(String arg, String[] args) {
+    return Stream.concat(Stream.of(arg), Stream.of(args)).collect(Collectors.toList());
+  }
+
+  private List<String> concat(List<String> a, List<String> b) {
```
achew22: This should be static

java/build/bazel/tests/integration/WorkspaceDriver.java
```diff
@@ -261,13 +270,77 @@ private BazelWorkspaceDriverException(String message) {
...
+        "--define=ABSOLUTE_JAVABASE="+ javaHome
+    );
+  }
+
+  private BazelCommand.Builder bazel(List<String> args, boolean addJavaBaseConfigFlags) {
+    return new BazelCommand.Builder(this, tmp, repositoryCache,
+        concat(args, maybeJavaBaseConfigFlags(addJavaBaseConfigFlags)));
+  }
+
+  private List<String> concat(String arg, String[] args) {
```
achew22: This should be static

tools/bazel_java_integration_test.bzl
achew22: +1

java/build/bazel/tests/integration/WorkspaceDriver.java
ittaiz: Javadoc? Readme? Both? I don’t like these methods but I’m not sure where else since this happens pretty implicitly to the user so it’s not trivial to propagate it back up in the method name

ittaiz:
> I'm starting to become even more convinced that we should move as much logic into .bzl as possible. Are there any major pitfalls you can see coming from that?
How do you see this working out? What parts?
The main two reasons I'm reluctant to make this kind of change are because there is a lot of value in declaratively controlling the bazel invocation from the test case (it's part of the test story) and because I think that setting up the different permutations outside will be noisy and unclear (for example a test which needs to run query and build and then needs different sets of flags)

tools/bazel_java_integration_test.bzl
ittaiz: Thanks! Not sure that would be correct. I agree that’s what I’m using it *for* but
OTOH I’m not using a custom java runtime for the test but rather passing
the real java runtime into the test.

PR #125, opened 2019-04-25
Disable tests on ubuntu1804 that don't work with JDK11

philwo:
I have migrated ubuntu1804 to use OpenJDK 11, because that's the default JDK for that distribution.

During the migration I noticed that you have tests that fail with OpenJDK 11, so I recommend to
disable them until they're fixed.

For your reference, here's the log: https://buildkite.com/bazel/bazel-integration-testing/builds/554#d64d47cf-3e61-4a9e-a075-7c62e7fae27d

### DIFF
--- a/.bazelci/presubmit.yml
+++ b/.bazelci/presubmit.yml
@@ -2,28 +2,31 @@
 platforms:
   ubuntu1604:
     build_targets:
-    - "..."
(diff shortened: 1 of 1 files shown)


ittaiz:
I fixed master so that it supports jdk11

PR #127, opened 2019-04-27
global properties are private and accessible as method accessor

ittaiz:
removes this irritating deprecated usage warning that we added so that people won't use our global mutable state

### DIFF
--- a/java/build/bazel/tests/integration/WorkspaceDriver.java
+++ b/java/build/bazel/tests/integration/WorkspaceDriver.java
@@ -37,8 +37,6 @@ public class WorkspaceDriver {
   private static Path tmp;
   private static Map<String, Path> bazelVersions;
   private static Path runfileDirectory = Paths.get(System.getenv("TEST_SRCDIR"));
-  // TODO: Don't have them as writable package-private when we have a better way
-  // to share configuration across the WorkspaceDriver instances.
   private static Properties properties;
 
   private Path currentBazel = null;
--- a/java/build/bazel/tests/integration/WorkspaceDriver.java
+++ b/java/build/bazel/tests/integration/WorkspaceDriver.java
@@ -63,8 +63,9 @@ public class WorkspaceDriver {
     javaToolchain = javaToolchainFromProperties();
   }
 
-  public static Optional<Properties> globalBazelProperties() {
-    return Optional.ofNullable(properties);
+  public static Properties globalBazelProperties() {
+    return Optional.ofNullable(properties)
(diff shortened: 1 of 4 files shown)


ittaiz:
@achew22 on the one hand there is an obvious code duplication required from clients which need the properties and can potentially get them from `setupClass`.
On the other hand `setupClass` is a `command` and not a `query` it's less appropriate for it to return a value. 
Additionally I think these properties will mainly (if not solely) will be used by this library to assert itself (like it's currently used).

This is why I think the current change is better than the existing code or having `setupClass` return the properties or a different struct.

https://stackoverflow.com/questions/28382151/command-query-separation-commands-must-return-void

ittaiz:
Thought of a much simpler solution than code duplication. I'm just asserting that `properties` is not null and if it is I'm throwing an exception clarifying `setupClass` must be called first

java/build/bazel/tests/integration/WorkspaceDriver.java
```diff
@@ -64,6 +63,10 @@ public static void setUpClass() throws IOException {
     javaToolchain = javaToolchainFromProperties();
   }
 
+  public static Properties globalBazelProperties() {
+    return Optional.ofNullable(properties)
+        .orElseThrow(() -> new IllegalStateException("Need to call setupClass() before calling globalBazelProperties()"));
```
achew22: Can you just call setupClass for them instead of throwing and make setupClass be idempotent? You could use a private static "alreadyInitialized" to achieve that.

ittaiz: I can but I don’t think it’s good design.
We’re not talking about runtime of web micro services. Worst case their test will fail, they will fix it and continue.
Why do you think implicit is better?

achew22: `setUpClass` takes no arguments and therefore has no configuration. You're throwing an exception that tells the user to call a method. Why put that on the user when you have already detected that the method wasn't called? Your API can have one function call instead of two, why not opt for one?

ittaiz: Because it’s very implicit that doing a “get” will trigger a big, IO including, setup.
https://stackoverflow.com/questions/28382151/command-query-separation-commands-must-return-void

Also remember that my intuition is that no actual user will need this getter. Maybe we should make it package protected.

achew22: If there is only a small chance it'll be needed let's make it protected with a comment

ittaiz: done

PR #131, opened 2019-05-01
ease test troubleshooting out of the box

ittaiz:
Global context:
I'm trying to use this library on a new repo and I'm stumbling on some caveats which we solved in our existing repo by wrapping this library.
I'm trying now to avoid this and actually upstream the fixes here.

Specific context:
Without this whenever a test fails it's very hard to understand what's going on without adding interim debug prints (interim because they are ugly) or debugging the test.

run command prints the toString of BazelCommand
added an opt-out runQuietly method

@or-shachar and @anchlovi I'd appreciate your review

### DIFF
--- a/java/build/bazel/tests/integration/BazelCommand.java
+++ b/java/build/bazel/tests/integration/BazelCommand.java
@@ -175,7 +175,7 @@ public class BazelCommand {
      * and returns an object to inspect the invocation result.
      */
     public BazelCommand mustRunAndReturnExitCode(int exitCode) throws IOException, InterruptedException {
-      BazelCommand cmd = run();
+      BazelCommand cmd = runQuietly();
       if (cmd.exitCode() != exitCode) {
         throw new RuntimeException(cmd + "==> exit code != " + exitCode);
       }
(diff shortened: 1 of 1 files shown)


ittaiz:
cc @achew22

anchlovi:
LGTM, though I would consider removing the `toString` and stream the output. This way the user will not need to wait for the command to complete its `run` to get the output

ittaiz:
@anchlovi thanks!
Re streaming- it’s an interesting idea. Two concerns there:
1. Clearly separating error lines from out lines. Separation we have now which I think is valuable.
2. We often need the out/error for matching purposes and this means we’ll need some kind of elaborate mechanism that streams to System.out while also keeping a copy like today. Sounds like complexity which I’m not sure is worth the value (which I agree exists).
Wdyt?
   
Additionally I’m actually having second thoughts about this change since maybe I should have just used mustRunSuccessfully and that’s it. This leads me to think maybe run should remain as is (quiet) and I add another method which is verbose (runVerbosely?).
Wdyt about that?

anchlovi:
>Additionally I’m actually having second thoughts about this change since maybe I should have just used mustRunSuccessfully and that’s it. This leads me to think maybe run should remain as is (quiet) and I add another method which is verbose (runVerbosely?).
Wdyt about that?

This sounds much better

ittaiz:
@anchlovi PTAL

PR #137, opened 2019-05-20
Windows: tests work with native test wrapper

laszlocsomor:
See https://github.com/bazelbuild/bazel/issues/6622

### DIFF
--- a/java/README.md
+++ b/java/README.md
@@ -51,5 +51,5 @@ public class ExampleTest {
 ```  
 In your test you should work with the `WorkspaceDriver` which has many useful utility methods for working with scratch workspaces.  
 These include, but are not limited to, `copyFromRunfiles`/`scratchFile`/`workspaceDirectoryContents`(useful for verbose error messages) and of course `bazelCommand` which allows you to run `bazel build` / `bazel test` / `bazel query` and more.  
-The [WorkspaceDriverIntegrationTest](../javatests/build/bazel/tests/integration/WorkspaceDriverIntegrationTest.java) has a few examples of realistic tests which sets up the environment and asserts Bazel's outputs (exit code, stderr, etc).  
-[WorkspaceDriverTest](../javatests/build/bazel/tests/integration/WorkspaceDriverTest.java) and [BazelBaseTestCaseTest](../javatests/build/bazel/tests/integration/BazelBaseTestCaseTest.java) show some more of the features.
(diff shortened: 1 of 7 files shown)


laszlocsomor:
("WIP" until I fix all tests.)

laszlocsomor:
Tests were failing because the inner Bazel's `ABSOLUTE_JAVABASE` path was near the `MAX_PATH` (260) limit on Windows. The tests classes' names are part of those paths, so renaming the test classes to a shorter name is one (albeit ugly) solution.

I'll look for a nicer solution now.

laszlocsomor:
After a few hours of failed attempts, I gave up and just renamed the test classes to shorter names.

achew22:
@iirina is the biggest user of the java part of the repo. Could you pull this PR into your repo and say if it causes issues?

laszlocsomor:
ping @iirina

iirina:
The PR looks good to me and the tests pass, so feel free to continue with the review.

laszlocsomor:
ping -- @achew22 , @ittaiz : does this LGTY?

achew22:
@laszlocsomor it feels very strange to me that the classes are going to get abbreviated in order to satisfy the naming requirements of Windows. Is this only exposed because Bazel has prefixed a huge number of characters in the prefix of the file?

laszlocsomor:
Yes and yes. Also because the prefix is doubled: once for the outer and once for the inner Bazel.

There's no rush though -- contrary to my plans, I can't enable this feature in Bazel 0.27.

achew22:
@laszlocsomor thanks for the quick response. Are there any plans to alleviate the XX character limit in a more general way? My concerns here revolve around the fact that if you were depending on Bazel Integration Testing in your repo, you would end up with the additional prefix of `build_bazel_integration_testing` (31 characters) in the file path, and I don't think it is possible to remove 31 characters from the classpath in here without seriously impeding readability.

I'm trying it on `build.bazel.tests.integration.BazelBaseTestCase` (47 characters), I assume that the domain has to be valid and owned, so `build.bazel` is a must. Given that constraint even `build.bazel.t.i.BBTC` is 51 characters and the illegibility of that would even make a Golang developer cringe!

Maybe now is a good time to take a step back and decide if there is a better solution maybe. I would be happy to schedule a meeting to talk it over with you internally, my @google.com email should be pretty easy to figure out and I'm happy to meet during EMEA friendly times.

PS: Thanks so much for caring about this project and putting in so much effort, I really appreciate it.

laszlocsomor:
To follow up, this PR is now obsolete: turns out the bug was in the test wrapper. See https://github.com/bazelbuild/bazel-integration-testing/issues/148

PR #149, opened 2019-08-03
Add alternate url for bazel binaries

gregmagolan:
release.bazel.build server went down for a while today while I was working on https://github.com/bazelbuild/rules_nodejs/pull/952 so I added this to get around the problem

### DIFF
--- a/tools/repositories.bzl
+++ b/tools/repositories.bzl
@@ -15,7 +15,10 @@
 load("//tools:bazel_hash_dict.bzl", "BAZEL_HASH_DICT")
 load(":common.bzl", "BAZEL_VERSIONS")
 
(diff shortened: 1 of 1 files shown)


tools/repositories.bzl
```diff
@@ -16,6 +16,7 @@ load("//tools:bazel_hash_dict.bzl", "BAZEL_HASH_DICT")
 load(":common.bzl", "BAZEL_VERSIONS")
 
 _BAZEL_BINARY_PACKAGE = "http://releases.bazel.build/{version}/release/bazel-{version}{installer}-{platform}.{extension}"
```
achew22: Could you convert this to a list of strings and then map the .format operation over the list to create the url list?

gregmagolan: 👍 good call

achew22:
@gregmagolan thanks so much for your contribution!

PR #152, opened 2019-09-09
Improve jvm readme

borkaehw:
We should provide users an easy setup example. By simply copy and paste, things should work.

I found some typos and possible improvements.


achew22:
@ittaiz this looks good to me, could you confirm this aligns with your understanding?

ittaiz:
Taking a look

PR #154, opened 2019-09-18
Fix failed JVM ExampleTest

borkaehw:
It turns out https://github.com/bazelbuild/bazel-integration-testing/pull/152 didn't get the example test fully functioning. I can confirm this PR gets it working.

Test is failing because `echo "boom"` goes to test log. I don't see a clear way to read test log content. I got around it by specifying [--test_output](https://docs.bazel.build/versions/master/command-line-reference.html). It outputs test log contents to stdout.

It could also be
```
BazelCommand cmd = driver.bazel("test", "--test_output=all", "//foo:bar").run();
```

### DIFF
--- a/java/README.md
+++ b/java/README.md
@@ -46,6 +46,7 @@ public class ExampleTest {
 
   @Test
   public void bazelFailure() throws Exception {
+    driver.scratchFile(".bazelrc", "build --test_output=all");
     driver.scratchFile("foo/BUILD", "sh_test(name = \"bar\",\n" + "srcs = [\"bar.sh\"])");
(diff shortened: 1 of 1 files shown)


ittaiz:
How is it related to 153? The example is `sh_test` Is it because you actually used it with `java_test` which failed (hence `errorLines` working)?

borkaehw:
Without #152, Bazel command failed. I wasn't paying attention that even though it built but test failed. #153 was the immediate reason I saw in test log. After solving `Class not found`, the test is still failing, and this PR will fix it.

#153 is indeed not directly related to the change, I will remove it from description.

borkaehw:
Reason for the changes was added to description.

